PR #4, opened 2021-04-05
adding formats

apreshill:
do not merge me- looks like draft pull requests are not available for private repos without GitHub Team, GitHub Enterprise Cloud, and GitHub Enterprise Server. So sharing work here, but 🚨 WIP 🚨


jjallaire:
@apreshill I'm pretty sure we've incorporated everything in this PR here (https://quarto.org/docs/output-formats/html-themes.html). Closing, but feel free to open a new PR if there is something we are missing.

PR #16, opened 2021-07-01
use styles.css filename as initial css example

aronatkins:
mirrors the later examples specifying css in YAML.


aronatkins:
FYI: A `quarto render` on this branch is not producing updated HTML; chatting with @dragonstyle about the problem.

Issue #20, opened 2021-07-28
Adding alt text for images in documentation


spcanelon:
Hello, and thanks for your work on this! It looks like a **great** system and I'm excited to try it out.

I noticed the documentation includes instructions on [how to add alternative (alt) text](https://quarto.org/docs/authoring/figures-and-layout.html#alt-text) within Quarto -- thanks for that! 🙌🏽 

I also noticed there are some images within the documentation itself that don't have alt text, which makes the information they are meant to convey inaccessible to R users that use screen readers. Images without alt text are often read aloud by default as "image" which you can imagine isn't helpful when trying to learn how to use a package. I'm opening this issue to share some encouragement and some resources that I'm hoping can help 🙂 

I haven't checked the images on _all_ of the documentation pages, but here are a few examples of images that still need alt text:
- Figure 1 in [Welcome to Quarto](https://quarto.org/#hello-quarto)
- Data viz in [HTML Code Blocks: Folding Code](https://quarto.org/docs/output-formats/html-code.html#folding-code)
- Screenshots in [HTML Code Blocks: Code Tools](https://quarto.org/docs/output-formats/html-code.html#code-tools), [HTML Code Blocks: Appearance](https://quarto.org/docs/output-formats/html-code.html#appearance), and [HTML Code Blocks: Highlighting](https://quarto.org/docs/output-formats/html-code.html#highlighting)

To determine what kind of images need alt text and how to write alt text I recommend the following resources from W3 and WebAIM:
- [Images Concepts • Images • WAI Web Accessibility Tutorials](https://www.w3.org/WAI/tutorials/images/)
- [An alt Decision Tree • Images • WAI Web Accessibility Tutorials](https://www.w3.org/WAI/tutorials/images/decision-tree/)
- [WebAIM: Alternative Text](https://webaim.org/techniques/alttext/)

Specific to writing alt text for data visualizations:
- [Writing Alt Text for Data Visualization](https://medium.com/nightingale/writing-alt-text-for-data-visualization-2a218ef43f81) by Amy Cesal
- [Revealing Room for Improvement in Accessibility within a Social Media
Data Visualization Learning Community: What makes good alt text?](https://spcanelon.github.io/csvConf2021/slides/#31) by Liz Hare and myself

Examples of R package documentation with alt text added by the package author:
- {stacks}: [Release stacks 0.2.1 • tidymodels/stacks](https://github.com/tidymodels/stacks/releases/tag/v0.2.1) by Simon Couch
[...]

PR #23, opened 2021-08-08
Add alt text to figures

### LINKED ISSUE #22: Links in Getting Started section are broken
All of the links in this section seem to be pointed to the ".md" versions and not the ".html" ones: https://quarto.org/#getting-started

pommevilla:
Address #20. I think I got them all! I tried to follow the guidelines I found in the sources provided by @spcanelon. That means I generally erred on the side of being too descriptive, but I also (generally) tried to stick to talking about what the point of the image was. Also, if a version of the same figure showed up multiple times in a page, I shortened the general description in later alt texts and focused on the details/differences.

@jjallaire, @dragonstyle:
* Can we get a live preview of these changes through Netlify? Or should I just host it on a fork?
* I only committed the changes to the `qmd` themselves. How is the site built? Should I render it locally and commit those changes, as well?

Additionally:
* Fixes #22 - the paths were just missing the 'docs' prefix. 
* A few typos I ran into
* There were a couple of images missing at the bottom of [books/book-authoring](https://quarto.org/docs/books/book-authoring.html). I copied the images over from [output-formats/ms-word.html#templates](https://quarto.org/docs/output-formats/ms-word.html#templates) into `books/images` and fixed those links.
* The first line in the example in  [visual_editor/technical.qmd#equations](https://quarto.org/docs/visual-editor/technical.html#equations) is wrong - it should be "The solution to \sqrt{x} = 26 is x = 676". I pushed a fixed version of the image: 
image
Github is adding a ton of whitespace above the image for some reason, but it doesn't show up in the preview. I used the Dreamweaver theme to take the screenshot. That's the one you used, right @jjallaire?


jjallaire:
Great, thanks!

You should be rendering this locally to see how the content comes out. I'd expect that anywhere you changed an image inside a grid table it would destroy the table (as they need to be precisely aligned). You should check all images in tables though to make sure the table is preserved. One way to manage this is to make the edits inside the visual editor rather than in source mode (pandoc will then make sure the table is preserved, even if it needs to write it as raw html to preserve the lengthy columns).

Branch previews are setup for Netlify, but you still need to render locally to `_site` before pushing for them to show up.

jjallaire:
Thanks for this!!! I had to make only one tweak which was to remove `fig-alt` from tables (as for grid tables which unfortunately the visual editor has a habit of producing) this will actually maul the table. My change is here: https://github.com/quarto-dev/quarto-web/pull/23/commits/a8fdbadade2a85b009c2e66300b5acc9d452713e

spcanelon:
Thanks @pommevilla and @jjallaire! 🙌🏽  
Hopefully the issue with getting `fig-alt` and tables to work well together can get resolved soon! I can imagine many scenarios where a user will want to make tables with images in them that are accessible to everyone.

PR #39, opened 2021-11-13
Language subtag is pr in the docs, but sould be pt (for Portuguese) (fix typo)

beatrizmilz:
Hi! In the website https://quarto.org/docs/authoring/language.html#translations the language subtag for Portuguese is `pr` but should be  `pt` (as it is in https://r12a.github.io/app-subtags/ and used in https://github.com/quarto-dev/quarto-cli/tree/main/src/resources/language )


beatrizmilz:
Sorry! I sent the PR quickly when I noticed the typo but did not give any context. Just edited so is more clear. :)

PR #66, opened 2022-02-08
remove duplicate entry for link-external-icon

chendaniely:
there are duplicate entries in https://quarto.org/docs/reference/formats/html.html#links

image

Note: I did not try to re-render the site in this PR. I just fixed up the `json` file.

### DIFF
--- a/docs/reference/formats/html.json
+++ b/docs/reference/formats/html.json
@@ -351,10 +351,6 @@
         "description": "Show a special icon next to links that leave the current site."
       },
       {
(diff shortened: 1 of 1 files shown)


jjallaire:
Thanks for this! The JSON is also generated, so I made the fix in ultimate source of truth (back in the quarto-cli repo): https://github.com/quarto-dev/quarto-cli/commit/091cdcda91922c49e7ab83d3d2b4f0ccc9aa417a

PR #110, opened 2022-04-06
Add bare link syntax

cderv:
Following https://github.com/quarto-dev/quarto-cli/discussions/540 and because `autolink-bare-uris` is not activated by default in Quarto. 

`autolink-bare-uris` is activate by default in R Markdown, so this will be part of the difference in rendering. 

This PR only have the `md` file change for this. I did not rebuild the website locally as I am not sure of the version used already, and the process (using RStudio visual editor or VS CODE - this change the syntax of some tables for example using the former)

Also on side note: 
Do we want a workflow to build the documentation on Github Action instead of rendering locally ? This would probably mean having `_site` not commited. Not sure how `_freeze` should be dealt with - probably cached in github action ? 🤔 
A good example to improve how actions documentation. I'll have a look anyway.


jjallaire:
We should in general use GH actions for building sites however for this particular site we still find it useful to render locally and observe diffs in `_site` output as a sort of regression check. So we could definitely add the GH action but we'd likely still check in `_site` for this purpose.

In terms of `_freeze` I think this always needs to be checked in to version control. If you merely cache it in the GH action then if this cache is at some point lost then you may be left unable to render those documents.

cderv:
> So we could definitely add the GH action but we'd likely still check in _site for this purpose.

This could be done somehow by having the action build the site for us based on the file changed and commit `_site` files back into the PR so that we could check before merging. 

> In terms of _freeze I think this always needs to be checked in to version control. If you merely cache it in the GH action then if this cache is at some point lost then you may be left unable to render those documents.

That is what I fear. This means that we need specific documentation on this, and maybe helpers as this is not straightforward. 
If you want to use `freeze`, either you render locally and push (as done in this repo), at least the document requiring a freeze, either you need  to have the github action commit the freeze folder into the repo. 

I'll look closer to that for our examples. This is the type of specificity that could require adding more actions in `quarto-dev/quarto-actions`

PR #124, opened 2022-04-14
Add solution to "No such kernel" from conda environment

rgaiacs:
Close https://github.com/quarto-dev/quarto-cli/issues/636


jjallaire:
Hi there, thanks for this! I've update the docs in a similar fashion to what you propose here: https://quarto.org/docs/computations/jupyter-kernels.html#kernel-selection

The main difference is that I link to the full instructions provided by nb_conda_kernels (as they contain additional details and will presumably be updated if anything changes).

PR #148, opened 2022-04-29
Mention that knitr also supports languages besides R

yihui:
Is it worth mentioning here, or would it be confusing (both knitr and jupyter support multiple languages, and users may naturally ask how to choose between which engine to use if they want to use multiple languages in the same doc)?


jjallaire:
Yes, I think this is indeed good to mention. You are right though that since both Knitr and Jupyter support R, Python, and Julia, if we say no more than that it could be confusing. Perhaps I should just say that R users will typically choose Knitr and Python/Julia users will typically use Jupyter?

jjallaire:
I just ended up adding R to the list of languages supported by Jupyter and didn't add any additional guidance about which engine to choose (I think this could get pretty nuanced/involved and elsewhere on the site we already nudge R users to Knitr and Python users to Jupyter).

Issue #154, opened 2022-05-06
Improve example of figures produced by executable code blocks to include subcaption


rgaiacs:
The example at https://quarto.org/docs/authoring/figures.html#figures doesn't demonstrate the use of fig-subcap.

PR #155, opened 2022-05-06
Add fig-subcap to example of figures produced by executable code blocks

rgaiacs:
Close https://github.com/quarto-dev/quarto-web/issues/154.

``````
```{r}
#| layout-ncol: 2
#| fig-cap: Main
#| fig-subcap: ["Foo", "Bar"]

plot(cars)
plot(pressure)
```
``````

generates

Screenshot from 2022-05-06 13-45-54

# TODO

I would love a little of mentoring to

- [ ] Fix the caption for the main figure that is missing in the output
- [ ] Update the generated PNG

### DIFF
--- a/docs/authoring/figures.qmd
+++ b/docs/authoring/figures.qmd
@@ -336,6 +336,8 @@ Note that figure layout attributes also work for figures produced by executable 
 #### Jupyter
 
 ``` {{python}}
+#| fig-cap: Charts.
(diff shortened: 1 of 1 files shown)


jjallaire:
I ended up just adding a new section (to demonstrate using layouts both with and without captions): https://quarto.org/docs/authoring/figures.html#figure-captions

PR #156, opened 2022-05-06
Update doc on revealjs plugin

cderv:
Follows https://github.com/quarto-dev/quarto-cli/issues/806

* correction option name for revealjs-plugin
* Add example based on the fullscreen plugin 

If the example is too much for the doc, maybe I could transform to a blog post ? 
I see there is a blog already (https://quarto.org/docs/blog/) and maybe this type of content is better there ?


jjallaire:
I think its great to have the full example here (no need to make it a blog post IMO)

PR #209, opened 2022-06-06
`linestretch` option must be a string instead of a numeric

pat-s:
fixes https://github.com/quarto-dev/quarto-cli/issues/1081 (sorry, opened issue in the wrong repo)


jjallaire:
@dragonstyle It seems like the schema should allow for either numbers or strings here — does it?

dragonstyle:
There was a version of this bug in the other repo which I fixed to allow either as you suggest. We can close this.

PR #210, opened 2022-06-07
Add cross-reference example table from code cell output

### LINKED ISSUE #206: Add example of cross-referencing a table with `tbl-cap:` in code chunk
Not a big deal, but took me a second to figure it out that I could use `#| tbl-cap: "Caption text {#tbl-knitr}"` to make the table output of a code chunk cross-referenceable, since there are no examples of this in the docs. 

I'm guessing this is because the cross-reference section is an `.md` include
https://github.com/quarto-dev/quarto-web/blob/c048cac782767f9efba6444188bf710d044d9868/docs/authoring/tables.qmd#L101-L106

So, I was thinking one solution would be to add something to the effect of `{#tbl-planets}` to the `tbl-cap` for the tabulate example in [Computations](https://quarto.org/docs/authoring/tables.html#computations) (since it's the chunk that's actually rendered). That way, you could quite literally cross-reference it in [Cross References](https://quarto.org/docs/authoring/tables.html#cross-references), and provide an example of its use without adding another table.
https://github.com/quarto-dev/quarto-web/blob/c048cac782767f9efba6444188bf710d044d9868/docs/authoring/tables.qmd#L133-L147

If this sounds good to you, I'm happy to submit a PR.

batpigandme:
* Closes #206


jjallaire:
Thanks! While it may work this isn't quite the recommended form though. Rather, it's to add a separate `label` option with the table identifier, e.g.

```yaml
label: tbl-planets
tbl-cap: "Planets"
```

PR #264, opened 2022-07-15
Gallery update

mine-cetinkaya-rundel:
- Add new presentations, websites, and books
- Split websites and books

Waiting on a couple more "yes"s from folks whose work we're including, will convert to ready to review when I get those.


jjallaire:
I'd say let's just remove the folks that haven't yet said yes and merge the PR (then just add the stragglers as they come in)

mine-cetinkaya-rundel:
@jjallaire Removed a few (and added a few more I recently got confirmations for). This should be ready to review/merge now. I'll add more as they come in.

PR #268, opened 2022-07-17
Yaml intelligence rstudio

mine-cetinkaya-rundel:
Matches https://quarto.org/docs/tools/vscode.html#yaml-intelligence with RStudio IDE screenshots. I think this is a good place to add this section but happy to move it elsewhere in the doc.


mine-cetinkaya-rundel:
Not sure why some checks are not successful but deploy preview looks ok to me.

PR #305, opened 2022-08-03
Make style of ggplot commands consistent

rhagenson:
While reading through https://quarto.org/docs/computations/r.html I noticed a few style inconsistencies in how ggplot command were formatted.


rhagenson:
I am going to run a full review of ggplot command formatting later today. Please keep this PR open until then.

rhagenson:
Later became now so this is ready.

I used a general rule that all additions to a ggplot should be on a new line and indented by 2 spaces. This seemed to be the most used style across the documentation site. I left any single line statements such as `p <- p + geom_jitter()` alone as I personally think a line break here would be unnecessary.

PR #310, opened 2022-08-07
Make it clear to the reader how to change page properties in Word

rgaiacs:
To PDF produced by LaTeX, I can change the page by having

```
format:
  pdf:
    papersize: a4paper
```

in the YAML header. I was looking how to do the same with Word and it strikes me that the documentation doesn't clear states it. [Pandoc's manual](https://pandoc.org/MANUAL.html) says

> The contents of the reference docx are ignored, but its stylesheets and document properties (including margins, page size, header, and footer) are used in the new docx.

This PR add a paragraphs about changing document properties.

**Help needed!**

- [ ] Create correct screenshot. I don't MS Word in my machine.
- [ ] Replace wrong screenshot.

### DIFF
--- a/docs/output-formats/ms-word-templates.qmd
+++ b/docs/output-formats/ms-word-templates.qmd
@@ -5,7 +5,7 @@ format: html
 
 ## Using Templates
 
-If you want to customize the appearance of MS Word output, Pandoc supports a special type of template called a *reference document*. Here's an example of specifying a custom reference document for `docx`:
+If you want to customize the appearance (from font type to page size) of MS Word output, Pandoc supports a special type of template called a *reference document*. Here's an example of specifying a custom reference document for `docx`:
 
 ``` yaml
 format:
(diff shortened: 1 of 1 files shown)


docs/output-formats/ms-word-templates.qmd
```diff
@@ -26,6 +26,10 @@ To create a new reference doc based on the Pandoc default, execute the following
 
 Then, open `custom-reference-doc.docx` in MS Word and modify styles as you wish:
 
+![You can open the Page pane from the HOME tab in the MS Word toolbar.](images/word-page.png){.preview-image fig-alt="Screenshot of Microsoft Word document open with Page  pane open in a pane over the left side of the document."}
```
jjallaire, line 29: This screenshot (images/word-page.png) also needs to be added to the PR.

rgaiacs: @jjallaire Thanks for the review. I don't have MS Word in my machine. Can you get the screenshot for me? Thanks!

PR #312, opened 2022-08-08
Citable field docs (only valid on pre-release branch)

dragonstyle:
Don't merge just yet please, but open so I don't forget about it!


jjallaire:
My thought was that we would indeed wait for v1.1 to merge docs but for a bigger item like this it is okay to merge and couch as pre-release only

dragonstyle:
Ok I'll add a callout that it is prerelease and merge it!

PR #333, opened 2022-08-15
Arrow -> Apache Arrow IPC

eitsupi:
The relationship between Apache Arrow and Feather is very confusing to the uninitiated, so I think it is better to use the term "Apache Arrow IPC file format" as used in the official documentation and push Feather V2 into parentheses.
(c.f. https://issues.apache.org/jira/browse/ARROW-17092)

Related to quarto-dev/quarto-cli#1845

### DIFF
--- a/docs/interactive/ojs/data-sources.qmd
+++ b/docs/interactive/ojs/data-sources.qmd
@@ -45,17 +45,18 @@ Note that we specified the `typed: true` option to the `csv()` function. When th
 
 Here are the methods available for structured data formats:
 
-| Method                                                                                                                       | Description               |
-|--------------------------------------------------|----------------------|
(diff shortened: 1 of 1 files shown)


eitsupi:
Hi, unfortunately, this change does not seem to be visibly reflected on the web site.
I'm wondering if it is possible for you to re-render the site.

eitsupi:
Seems like this file needs to be updated? (Sorry, I don't fully understand how freeze works)
https://github.com/quarto-dev/quarto-web/blob/bb90419c495ca79789cf6a74fe43bd2f63cebf8b/_freeze/docs/interactive/ojs/data-sources/execute-results/html.json#L4

jjallaire:
Thanks, done!

PR #400, opened 2022-09-19
Bump Quarto Version indicated for `quarto.log.output`

hamelsmu:
I tried with the version currently in the docs, and it doesn't seem to work.  However, the latest pre-release does work.


jjallaire:
Thanks! Made this same update elsewhere so closing w/o merging.

PR #431, opened 2022-10-07
Explain how to use with python virtual environments

SamEdwardes:
Hi Quarto team,

I was recently using Quarto to document a Python package I am developing. It took me some time to figure out how to get Quarto to use my virtual environment. Eventually, I realized I needed to install a kernel within my virtual environment. Documenting how to do this could be helpful as other users may run into the same problem.

I have added a new section to the docs/computations/python.qmd page to explain how to use Quarto with a virtual environment.

Note that it is already documented how to select your kernel:

https://github.com/quarto-dev/quarto-web/blob/81540c76b5c84be6e3e5f402c7796e55de58f6a5/docs/computations/python.qmd#L83-L105

I hope that my PR makes it clear to users that you need to install a kernel in your virtual environment.

### DIFF
--- a/docs/computations/python.qmd
+++ b/docs/computations/python.qmd
@@ -116,6 +116,67 @@ If you are using a kernel that is contained within an external conda environment
 Note that this step is not required if you are merely using conda with Quarto. It applies to using kernels other than the default Python kernel that happen to be installed within a conda environment separate from the one you are using.
 :::
 
+## Kernels with Virtual Environments
+
+To use Quarto with a virtual environment, you need to install a kernel within the virtual environment. Follow these instructions to create a new virtual environment and install a kernel.
+
+:::{.panel-tabset}
+
+## Windows
+
+```{.bash filename="Terminal"}
+# Create a new virtual environment
+python -m venv .venv
+
+# Activate the virtual environment
+.venv\Scripts\activate.bat
+
+# Install ipykernel
+python3 -m pip install jupyter
+
+# Install a new kernel
+python3 -m ipykernel install --name my-project-kernel --display-name "Python Kernel for My Project" --user
+```
+
+## Mac/Linux
+
+```{.bash filename="Terminal"}
+# Create a new virtual environment
+python -m venv .venv
+
+# Activate the virtual environment
+source .venv/bin/activate
+
+# Install ipykernel
+python3 -m pip install jupyter
+
+# Install a new kernel
+python3 -m ipykernel install --name my-project-kernel --display-name "Python Kernel for My Project" --user
+```
+
+:::
+
+Then, in your Quarto document, reference the new kernel you installed by name:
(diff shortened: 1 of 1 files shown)


jjallaire:
My reservation about including this documentation as-is is that it is generally NOT required to specially install a kernel when using a virtual environment (as `pip3 install jupyter` installs everything that is required by Quarto). I've never seen anything more than this required and users have not reported any inability to use virtual environments without this extra step. i.e. I have never needed to do this on any platform nor have I ever seen a user need to do this:

```bash
python3 -m ipykernel install --name my-project-kernel --display-name "Python Kernel for My Project" --user
```

I am also able to use all of the kernels registered on my system from within a Python virtual environment without any special installation or registration. Here is the output from `quarto check` when running inside a freshly created Python venv:

```bash
[✓] Checking Python 3 installation....OK
      Version: 3.9.5
      Path: /Users/jjallaire/quarto/dev/python-venv-project/env/bin/python3
      Jupyter: 4.11.1
      Kernels: ir, julia-1.7, bash, julia-1.5, python3, c
```

I am able to use these kernels in my venv without any additional installation steps. This is because generally speaking Jupyter kernels are installed globally (e.g. you don't need to install the bash kernel over and over again into every venv that you create). 

It does seem to be true that if you want to use a _custom_ Python kernel of some kind (i.e. not just the default Python kernel) that you do need to install it explicitly into virtual environments. 

I think if we want to include this documentation we need to extremely clear that this is _only_ for that very specialized case. Even then, if a user is off creating virtual environments and using custom kernels that is quite advanced and they really should be able to sort things out (i.e. I am wary of cluttering up our documentation with advanced use cases lest we actually lead more mainstream users into that quagmire). For this reason I'm inclined to not include this unless we believe specifically that custom Python3 kernels are a somewhat mainstream requirement.

SamEdwardes:
Thank you for the detailed reply JJ. I think I have misunderstood how Quarto works with virtual environments. I was running into all sorts of issues, and for some reason installing the kernel appeared to fix it.

I will see if I can create a reprex and share it. In the meantime, I agree we should not merge my proposed changes to the docs.

SamEdwardes:
Issue created here: https://github.com/quarto-dev/quarto-cli/issues/2801

paniterka:
I feel that this is more a Python/Jupyter maintenance issue than a Quarto problem. Generally there is no need to install full jupyter in every virtual environment - it is better to have one environment (such as base) running jupyter server and have all the other environments have only their ipython kernels registered in this one jupyter instance (precisely by activating the environment and running `python3 -m ipykernel install --name my-project-kernel --display-name "Python Kernel for My Project" --user`). This is a setup I am using and when I list the available jupyter kernels with `jupyter kernelspec list`, I see the same available kernels as when using `quarto check`.

PR #442, opened 2022-10-13
change rmd to qmd

Ifitrains:
I would also opt for using the same example qmd also in the `quarto::quarto_render` example.


dragonstyle:
Thank you!

PR #447, opened 2022-10-15
Quarto executable source code, without executing it

aborruso:
How to add it

### DIFF
--- a/docs/output-formats/html-code.qmd
+++ b/docs/output-formats/html-code.qmd
@@ -366,3 +366,53 @@ format:
   html:
     code-line-numbers: true
 ```
+
+## Insert Quarto executable source code, without executing it
+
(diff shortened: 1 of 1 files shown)


jjallaire:
We have this documented over here already: https://quarto.org/docs/computations/execution-options.html#unexecuted-blocks

Perhaps we should just duplicate that explanation in the spot you are pointing out?

aborruso:
@jjallaire you are right. I had looked for it, but I had not found it.

Because for me this is an authoring matter and not an execution one. 
I was looking for it in the sections that explain how to write content.
I would also put a reference there.

What do you think about it?

Thank you

jjallaire:
Okay, I just pushed a change that includes this documentation in both places so will close the PR.

aborruso:
Thank you very much

PR #450, opened 2022-10-17
Update journal-articles.yml

jchiquet:
Added an entry for the template (work-in-progress) of the [Computo journal)(https://computo.sfds.asso.fr/).


jchiquet:
my pleasure, thanks.

Issue #422, opened 2022-09-30
Update doc to use `embed-resources` instead of `self-contained`


cderv:
Now that it is deprecated in Pandoc, we should recommend using `embed-resources`. Just opening the issue to track the need to do it in all docs.

PR #462, opened 2022-10-25
fix: uses `embed-resources` instead of `self-contained`

mcanouil:
This PR fix the last file where `self-contained` was still mentionned (inside a callout).
See <https://quarto.org/docs/presentations/revealjs/presenting.html#chalkboard>.

With this PR, #422 can be closed I think.

To note, there is one file in `_freeze` still mentionning the deprecated option: <https://github.com/quarto-dev/quarto-web/tree/main/_freeze/site_libs/revealjs/plugin/reveal-chalkboard/plugin.yml>


dragonstyle:
Thank you!

PR #464, opened 2022-10-25
Add RSS icon

aborruso:
Moreover it's a way to show a quarto feature


dragonstyle:
Thank you for this PR! I ended up making the change per your suggestion but including some additional CSS changes to make the appearance optimal. Commit here:

https://github.com/quarto-dev/quarto-web/commit/f90241f5a6e48e92c554425cf829610b2b05774e

PR #470, opened 2022-10-28
add clarification for relative file locations

jlgraves-ubc:
This hopefully will save other people time, if you're dealing with complex websites!


jlgraves-ubc:
I will mention I'm actually not sure if this functionality is intentional or not!

PR #487, opened 2022-11-05
add my extension

jlgraves-ubc:
It's to create forms!  Why do you need forms in static sites?  Don't ask!


jjallaire:
Thanks!!!!

PR #489, opened 2022-11-10
Add link to a list of supported syntax highlighting languages

owiecc:
Nice to know which languages are actually supported and where to propose additions.


dragonstyle:
Thank you!

PR #516, opened 2022-11-25
add extension: code-insertion.

feynlee:
Add extension `code-insertion` to add markdown/html code immediately before or after the post.


dragonstyle:
Thank you for the PR! Before we merge this, could you help us understand the difference between what this extension provides vs our built in `include` or `include-in-header` type functionality?

dragonstyle:
OK great, thanks for the clarification. Definitely sounds like a useful extension!

PR #522, opened 2022-11-30
Add confetti extension

ArthurData:
I have just created this extension. I would like to add it to the list of quarto extensions. 
Thanks :)


jjallaire:
Thanks!

PR #526, opened 2022-12-09
documentation: use fig-alt instead of fig-cap for Multiple plot examples

zkamvar:
I would suggest using `fig-alt` as an example for demonstrating Multiple Outputs as a more practical (albeit less visible for sighted users) and instructive alternative to `fig-cap`.

By showing it prominently in the docs, the practice of including accessible alt text will be more normalised. My attempts are here, but feel free to provide a better description if you are better at describing curves than I am.

### DIFF
--- a/docs/computations/julia.qmd
+++ b/docs/computations/julia.qmd
@@ -59,14 +59,14 @@ There are many options which control the behavior of code execution and output, 
 
 ### Multiple Outputs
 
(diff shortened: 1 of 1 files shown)


jjallaire:
For this particular example I prefer that we keep with `fig-cap` (as "how do I do numbered sub-captions" is a very frequent question so I want to keep in prominent in the overview doc). Would however be very welcome to an additional example demonstrating fig-alt.

PR #550, opened 2023-01-04
Instructions to create an empty gh-pages branch

jmp75:
While migrating [this documentation repository](https://github.com/csiro-hydroinformatics/streamflow-forecasting-tools-onboard/) to use Quarto, I am bumping into a bit of a slightly confusing step in the `Publish Command` section. Instructions state that:

> Before attempting to publish you should ensure that the **Source** branch for your repository is `gh-pages`

but typically a repository would not have a pre-existing `gh-pages` branch.

I submit a few command lines to set up an empty stub `gh-pages` branch.

### DIFF
--- a/docs/publishing/github-pages.qmd
+++ b/docs/publishing/github-pages.qmd
@@ -64,9 +64,20 @@ The `quarto publish` command is an easy way to publish locally rendered document
 
 ### Source Branch {#source-branch}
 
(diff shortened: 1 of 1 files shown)


jjallaire:
Interesting! GH appears to have changed this UX (they used to allow you to set this even if you didn't have a `gh-pages` branch -- now it looks like b/c they allow publishing from any branch they got rid of the pre-stocked `gh-pages`).

PR #556, opened 2023-01-09
reference template partials

joelvonrotz:
When searching for example for "quarto custom title page" I get the title blocks page, which explains on how to customize the blocks. I think a guide or reference on how to change specific parts of a document, one of it being the title page/block could be useful for others.

Any thoughts on it?


dragonstyle:
Thank you for this! I've made a slightly tweaked version of this (with a relative URL) here:

https://github.com/quarto-dev/quarto-web/commit/a4823f783e525281209756cd01e934ebe3e03f26

joelvonrotz:
Thank you! My English is not the best, so your tweaked version reads a lot better 😄

Issue #566, opened 2023-01-14
Duplicat entries in books.json


jmcastagnetto:
The file https://github.com/quarto-dev/quarto-web/blob/50b8e37ab46690f04be9d96a5955928132810449/docs/reference/projects/book.json, contains duplicate entries, some that do not seem to correspond to the quarto book type.

I am submitting a quick PR to fix this issue.

dragonstyle:
Thank you for noticing this error! This file is generated using our underlying schema for books, so the duplication points to an issue that needs to be fixed there. I believe in this case the likely issue is overlap between our `book` schema in Quarto and the `CSL` schema which the book also provides. I'm leaving this issue open so @cscheid and I can track this down, I'm not sure the best thing to do in the event of these type of conflicts, so we'll want to talk it through. 

I'll close the PR for now (thank you, though), as we don't want to change that file manually since those change will be lost the next time we regenerate the references docs using the schema.

PR #567, opened 2023-01-14
Removing duplicate entries

jmcastagnetto:
Some entries were duplicated, and in one case ('references') it had a second definition that did not correspond to the use.

This PR will fix issue #566.


dragonstyle:
Thanks for the PR, see comments on https://github.com/quarto-dev/quarto-web/issues/566 for an explanation - I think we need to fix a deeper issue to resolve this!

PR #574, opened 2023-01-18
Added guide on how to enable snippets suggestion

joelvonrotz:
I was looking for a way to enable snippet suggestions in `qmd` files and found a stackoverflow post that explained it. I can't find it anymore though, but I'd like to add a quick guide on how to enable it. Also it seems that the Quarto extension overrides the default setting.

### DIFF
--- a/docs/tools/vscode.qmd
+++ b/docs/tools/vscode.qmd
@@ -143,6 +143,14 @@ Code snippets are templates that make it easier to enter repeating code patterns
 
 ![](images/vscode-snippets.png){.border fig-alt="Quarto document with dropdown 'Select a snippet' dropdown, the first item (bold - insert bold text) is selected."}
 
+### IntelliSense
+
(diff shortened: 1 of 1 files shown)


joelvonrotz:
This is what the setting looks like when all the snippetSuggestions settings are reset.

grafik

It would be nice to either have the guide or update the extension.

jjallaire:
Thanks for this! We override this setting because we also enable quick suggestions (i.e. always be looking for completions and don't wait for an explicit ctrl+space). When quick suggestions are enabled snippet suggestions end up showing up constantly in the normal course of typing (e.g. "bo" shows the `bold` snippet, "he" shows the `heading` snippets), which ends up being quite annoying!

If there was a way to have snippet suggestions _only_ show up on ctrl+space that would be ideal. I guess another approach entirely would be to have our snippets start with an infrequent character sequence (e.g. `>>`) which would make it dramatically less likely that they'd show up while typing prose.

joelvonrotz:
Thank you for the merge.
True, that is annoying and I hope some kind of conditional snippets solution can be developed by the vscode team.

PR #598, opened 2023-02-01
Added an entry for shortcode in shortcode-and-filter listing

shafayetShafee:
[`downloadthis`](https://github.com/shafayetShafee/downloadthis) provides shortcode for implementing download buttons in html output file.


jjallaire:
Thanks, looks great!

PR #603, opened 2023-02-03
added an entry for a filter in shortcode-and-filter listing

shafayetShafee:
Added the filter [`include-code-files`](https://github.com/shafayetShafee/include-code-files) that extends the [`include-code-files`](https://github.com/quarto-ext/include-code-files) by [@quarto-ext](https://github.com/quarto-ext), so that included code blocks can be folded with `code-fold: true` and also provided an option so that [`filename`](https://quarto.org/docs/output-formats/html-code.html#code-filename) works fine along with code-folding.


jjallaire:
I think it would be great if this had a slightly different name from the original just for easier top-level disambiguation

shafayetShafee:
How about `add-code-files`? Or can you suggest a good different name? I am very bad with naming :(

PR #628, opened 2023-02-17
Multi-format docs

cwickham:
Changes:

* Reorder and minor rewording
* Add note to get all formats rendered in RStudio and VS Code.

Fixes quarto-dev/quarto-cli#4453

### DIFF
--- a/docs/prerelease/1.3/multi-format.qmd
+++ b/docs/prerelease/1.3/multi-format.qmd
@@ -22,7 +22,7 @@ format:
   ipynb: default
 ```
 
-Results in an HTML page that includes a link to the additional notebook format in the right margin below the TOC:
+Results in an HTML page that includes a link to the additional notebook format in the right margin below the table of contents:
 
 ![](images/other-format.png){.border fig-alt="Screenshot of a HTML page that includes a link to the Jupyter format below the table of contents under the heading Other Formats."}
 
@@ -43,7 +43,7 @@ quarto render multi-format.qmd
 
 ## Rendering Formats with the Same Extension
 
-If your formats share a file extension, for example an HTML page and a Revealjs presentation both use `.html`, their outputs will overwrite each other. To include formats with the same extension, use `output-file` to distinguish one. For example, to include a link to a RevealJS presentation provide a distinct `output-file`:
+If your formats share a file extension, for example, both HTML pages and Revealjs presentations use `.html`, their outputs will overwrite each other. To include formats with the same extension, use `output-file` to distinguish one. For example, to include a link to a RevealJS presentation provide a distinct `output-file`:
 
 ```{.yaml filename=document.qmd}
 title: Sample Page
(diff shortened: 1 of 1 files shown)


mine-cetinkaya-rundel (approved):
This is great, thank you! I implemented it at https://vizdata.org/course-syllabus.html, which is one place I always need multi-formats.

One thing I tried first was to add just 

```
format:
  pdf: default
```

to the YAML of the qmd document for this page, but I then realized I need to specify both

```
format:
  html: default
  pdf: default
```

Even though `format: html` is already specified at the project level. So the metadata is not "additive", in a way. This might be good to specify at the end?

mine-cetinkaya-rundel:
Also, I wonder if it makes sense to mention this doesn't work for getting revealjs and pdf together?

cwickham:
> Also, I wonder if it makes sense to mention this doesn't work for getting revealjs and pdf together?

Do you mean this [issue](https://github.com/quarto-dev/quarto-cli/issues/4398)? I think it should work once that bug is fixed. Not sure, if I should mention it in these docs, or just delay publicizing until it is fixed.

dragonstyle:
> > Also, I wonder if it makes sense to mention this doesn't work for getting revealjs and pdf together?
> 
> Do you mean this [issue](https://github.com/quarto-dev/quarto-cli/issues/4398)? I think it should work once that bug is fixed. Not sure, if I should mention it in these docs, or just delay publicizing until it is fixed.

This one is going to be tricky to fix (the real issue isn't just the linking, it is that the html and revealjs formats currently overwrite one another since they are both outputting HTML files). It's possible we'll have to settle asking the user to set an `output-file` for one or the other (and being sure that the linking respects that output file, which it should).

cwickham:
@dragonstyle I think that is a reasonable solution - adding `output-file` was the first thing I tried, but it didn't work.

cwickham:
I'll add a section about adding formats that would otherwise overwrite each other.

docs/prerelease/1.3/multi-format.qmd
```diff
@@ -61,8 +84,21 @@ format:
...
+
+The `format`{spellcheck="false"} option isn't merged like all other [metadata](https://quarto.org/docs/projects/quarto-projects.html#metadata-merging) across `_quarto.yml`{spellcheck="false"}, `_metadata.yml`{spellcheck="false"}, and the document YAML. If you have some formats specified at a project or directory level, you'll also need to explicitly list them in the document YAML. For example, if you have HTML options set at the project level:
+
+``` {.yaml filename="_quarto.yml"}
+format:
+  html:
+    toc: true
+```
+
+To get a HTML document with a link to the PDF format, you'll need to list both formats in the YAML header:
```
mine-cetinkaya-rundel, line 99: Should this start with a lower case? Or maybe the "If ..." sentence above can be "Suppose you have HTML options set at the project level:" then it's ok that "To ..." is a new sentence.

mine-cetinkaya-rundel:
@cwickham re: comment about revealjs and pdf, I _think_ the following doesn't work:

```
format: 
  revealjs: default
  pdf: default
```

Maybe I'm mistaken? But when I try this I don't get both outputs.

If that's in case not a combo that works, might be worth mentioning it. Otherwise, everything looks great, I left just one more wording suggestion.

docs/prerelease/1.3/multi-format.qmd
cwickham: Yes, fixed.

PR #636, opened 2023-02-20
Confluence Docs

jjallaire:
This looks great! So excited to get this into people's hands :-) This PR makes some suggested revisions --- nearly all of the original content is still there (save for stuff about project workflow that isn't as applicable to Confluence) but I've added some introductory material and some info on using the "preview formats" for both documents and projects.

I would also suggest that for the time being we move this to be a peer of the other v1.3 preview feature documentation here: https://github.com/quarto-dev/quarto-web/tree/main/docs/prerelease/1.3 (we are no longer using the `{{< include >}}` from the publish directory which will make this easier).

Once we've iterated on this and are happy with the basic flow, I think we should add a screenshot and perhaps diagram at the beginning to sell the feature a bit more (perhaps a `qmd` on the left and a published confluence page that includes code, tables, etc. on the right?). Besides getting people excited about the feature, I think this will also help clarify in a more tactile way exactly what it is doing.

We should also work with @dragonstyle on getting a `quarto create website confluence` command added (and then perhaps split out "creating a new site" and "publishing an existing site" within the docs where the former uses this command.

In terms of diagrams, we might be able to get these working by just adding `mermaid-format: png` to the format when we render for confluence? I could definitely be wrong about this though (more on mermaid options here: https://quarto.org/docs/authoring/diagrams.html#mermaid-formats)


allenmanning (approved):
This looks great.

docs/prerelease/1.3/confluence.qmd
```diff
@@ -2,19 +2,184 @@
 title: "Confluence"
 provider: confluence
 provider-name: Confluence
-provider-publish-url: "https://myteam.atlassian.net/wiki/spaces/TEAMSPACE/pages/123456/Plan"
 ---
 
-{{< include _pre-release-feature.qmd >}}
+::: callout-note
```
allenmanning, line 7: @jjallaire As suggested, I've moved this to the pre-release section.

PR #639, opened 2023-02-22
Grid Customization Docs

cwickham:
Rewrite and reorder of Grid Customization Docs

Fixes quarto-dev/quarto-cli#4466


docs/prerelease/1.3/grid.qmd
```diff
@@ -4,15 +4,44 @@ title: Article Grid Customization
 
 {{< include _pre-release-feature.qmd >}}
 
-### Background
+## Overview
 
-Quarto HTML documents are arranged in a structure composed of a sidebar, the body of the document, the margin of the document, and the margins between these elements (gutters). This is illustrated below:
+Starting in Quarto v1.3, you can control the width of the layout components in HTML documents. For example, for a website that makes use of the margin for figures, it might make sense to give more space to the margin and less to the body:
```
dragonstyle, line 9: One other very common use case that I've heard (might be worth including as an example as well) is to increase the width of the sidebar (we have a pretty narrow default width and many people run into sidebars with longer titles that are wrapping and look sort of bad).

docs/prerelease/1.3/grid.qmd
```diff
@@ -4,15 +4,44 @@ title: Article Grid Customization
...
+
+This change can be made by adding the `grid` option to the `_quarto.yml` file, increasing the `margin-width` and decreasing the `body-width` from their defaults:
+
+```yaml
+format:
+  html:
+    grid:
+      margin-width: 450px
+      body-width: 600px
+```
```
dragonstyle, line 25: Is it worth noting that this change would keep the overall page width the same, allocating more space to margin.

Could also consider including an alternative that creates a larger margin, but leaves the body the same size, which would have the effect of making the overall page wider to accommodate the margin elements.

dragonstyle (approved):
I think this looks very good - a couple of suggestions for examples to consider including.

docs/prerelease/1.3/grid.qmd
mine-cetinkaya-rundel: Agreed, that would be nice. Or, at a minimum, mentioning the default page width would be good so it's clear this example is splitting it vs. it's possible to define `margin-width` and `body-width` that add up to more than that value.

docs/prerelease/1.3/grid.qmd
```diff
@@ -4,15 +4,44 @@ title: Article Grid Customization
...
+
+::: {.callout-note}
+## Websites vs. Standalone HTML Pages. 
+
+Customizing the layout of pages that are part of a Quarto website with YAML should happen at the site level in `_quarto.yml`. For HTML documents that aren't part of a website, these options could also be set in the YAML at the top of the document.
+
+:::
+
+There are four variables to control the four components of the layout: the sidebar, the body, the margin and the gutters.
```
mine-cetinkaya-rundel, line 34:
```suggestion
There are four variables to control the four components of the layout: the sidebar, the body, the margin, and the gutters. 
```

mine-cetinkaya-rundel (approved):
Looks great! Just left one comment.

docs/prerelease/1.3/grid.qmd
cwickham: That's a great example use case. I added it as the lead example, and moved the existing one later on the page.

docs/prerelease/1.3/grid.qmd
cwickham: Agreed, I've expanded this example to include both cases (increase overall page width, and keep overall page width the same)

PR #647, opened 2023-03-01
Extension list: add first-line-indent filter

jdutant:
Used in production for the journal Dialectica, sample: https://dialectica.philosophie.ch/dialectica/article/view/12


jjallaire:
This extension looks fantastic, thanks so much for listing it here!

PR #651, opened 2023-03-07
re-render docs

malcolmbarrett:
After https://github.com/quarto-dev/quarto-cli/pull/4657, although it looks like a lot else has changed since the last render.


malcolmbarrett:
No worries! I bet you can orchestrate this with GitHub actions where a certain type of release triggers a rebuild of the site

PR #655, opened 2023-03-09
Add how to specify multiple Figure params

DavidD003:
As a first time user reviewing the documentation and working with this kind of language, I had to experiment to find out how to apply multiple formatting specifications. I thought it was worth including in the documentation.

First time I do this kind of thing so please double check I didn't break anything else as I scrolled all over the page copying and pasting!


jjallaire:
Thanks!

PR #667, opened 2023-03-16
Add a specific workflow for deploying preview to netlify for PR

cderv:
This is a first try at running a preview on website for PR

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -1,0 +1,42 @@
+on:
+  workflow_dispatch:
+  pull_request:
+    branches: [main]
+
+name: Deploy Preview
+
+jobs:
+  build-deploy-preview:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v2 
+
+      - name: Set up Quarto
+        uses: quarto-dev/quarto-actions/setup@v2
+        with:
+          version: pre-release
+
+      - name: Render and Publish 
+        uses: quarto-dev/quarto-actions/render@v2
+        with: 
+          to: html
+
+      - name: Deploy Preview to Netlify as preview
+        id: netlify-deploy
+        uses: nwtgck/actions-netlify@v2
+        env:
+          NETLIFY_SITE_ID: 2a3da659-672b-4e5b-8785-e10ebf79a962
+          NETLIFY_AUTH_TOKEN: ${{ secrets.NETLIFY_AUTH_TOKEN }}
+        with:
+          publish-dir: './_site'
+          production-deploy: false
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          deploy-message:
+            'Deploy from GHA: ${{ github.event.pull_request.title || github.event.head_commit.message }} (${{ github.sha }})'
+          # these all default to 'true'
+          enable-pull-request-comment: true
+          enable-commit-comment: true
+          enable-commit-status: true
+          overwrites-pull-request-comment: true
+        timeout-minutes: 1


cderv:
@cwickham @allenmanning This PR adds a deploy preview for our website for each PR. 

Current situation: 

- It seems netlify website now lives in RStudio organisation under a non-explicit name so hard to find. I don't have access to this site. 
- The netlify app probably integrates with Github with the aim of having [deploy previews on PR](https://docs.netlify.com/site-deploys/deploy-previews/) - this should work but it does not **because site is not rendered with Quarto in the PR**. This means the build should happen on netlify, but this requires the [quarto netlify plugin](https://github.com/quarto-dev/netlify-plugin-quarto) to be configured, but I don't think it is for this website (can't check for sure as I don't have the right access to do it). 

So one fix would be to use our Netlify plugin, but I don't think it supports yet to build with pre-release version of Quarto, and we want that. 

So I used another approach, more usual that I know

- I used a specific action that triggers on PR only. 
- This action will render the website to HTML using quarto pre-release into the `_site` folder. 
- This folder is then deployed to netlify using a github action from marketplace, usual one we use in other project, **with specifying to deploy only for non-production** so deploy previews. 

This actions will add commit status with deployment URL
image

 and a Github environment status 
image

There is also this button it seems (not sure the parameter for that) 
image

So lots of ways to have the deployment url. 

It could also do PR comments like this one: https://github.com/quarto-dev/quarto-web/pull/667#issuecomment-1471696955 and overwrite it, but I figure it was not necessary so I changed the default in 7bb7bf85e7d876f2f70279bec4d0161afb81402b

Next step for this works: 

- [ ] Is this good solution for you ? Do we prefer to use the quarto netifly plugin for some reason ? probably needs an update though...
- [ ] Do you have adjustment @cwickham ? We can discuss anything you'd like.  
	-  about how the deployment information should be shown ? (Pr comment only ? Commit status only ? else ? ) 
	-  about the preview deployment url ?
- [ ] Who can give me access to our Netlify website so that we deactivate the integration for deploy preview ? 


Happy to demo and discuss any other improvement you need.

allenmanning:
I can't think of a reason why we would need to fix the older Netlify approach if this is working.  It also seems like this is easier to maintain for the reasons you listed above.

allenmanning (approved):
This looks great.  🚀 It also seems easier to maintain.

cderv:
Great ! Waiting for @cwickham thoughts before merging. I just would like to remove the check failure from Netlify - but I need access to the website configuration 😃

cwickham:
This looks fantastic! 
I think this PR comment is the most useful way to show the deployment (the URL is easy to find without any extra clicks):
Screen Shot 2023-03-17 at 9 19 16 AM

I agree with getting rid of those failing netlify checks - very distracting!

I'd also like to see a brief description of this process of PR previews end up in the (currently non-existent) README. I'll add a note on the issue for that, to keep track of it.

cwickham (approved):

cderv:
Thanks for the feedback @cwickham - I will reactivate the PR comments then. Should I remove one of the other integration if this is distracting ?

As soon as I have access to the netlify website, I'll remove the integration creating the checks. 

I'll probably merge this before doing the documentation. Will you initiate a README and I add to it after it is created ?

cwickham:
> Thanks for the feedback @cwickham - I will reactivate the PR comments then. Should I remove one of the other integration if this is distracting ?

If deployment isn't successful, one of the others might be useful...but without seeing a failing deploy it's hard to know which one (or even if both could go).  If we got rid of both, would there be any indication of a failed deploy? Maybe get rid of the "Github environment status" and we'll see how things work for a while.

> As soon as I have access to the netlify website, I'll remove the integration creating the checks.

👍🏻 

> I'll probably merge this before doing the documentation. Will you initiate a README and I add to it after it is created ?

👍🏻

cderv:
> I think this PR comment is the most useful way to show the deployment (the URL is easy to find without any extra clicks):

I made the change to reactivate the comment, but set it up to overwrite the previous one. See https://github.com/quarto-dev/quarto-web/pull/667#issuecomment-1471696955 that was edited. Is that useful as is or do you prefer a new comment at each deploy in the PR thread ¿

As you can see I also used the `deploy-preview-<PR number>--` for the URL now. Seems better than a random number.

> As soon as I have access to the netlify website, I'll remove the integration creating the checks.

So I have access now, and I have deactivated the the netlify integration for our site... but it seems the necessary action are not been done when moving to our netlify org with what was previously done. 

Currently website is still automatically deployed (but failling) at the olde website than @apreshill had setup at the early time of this website. So this repo is still connected to @apreshill [account wesbite](https://app.netlify.com/sites/vigilant-jennings-24cf11/deploys) 

@apreshill if you pass around here, is this possible to see to together to remove the website completely or at least disconnect the integration ? I have not manage to find the configuration from this Github repo itself. 

---------

Also, just putting that here for reference and discussion about why integration with netlify could be interesting for you @cwickham : Netlify has a specific contribution feature called Netlify Drawer to [collaborate on deploys](https://docs.netlify.com/site-deploys/collaborate-on-deploys/) - this is only available if we deploy the website using Netlify integration for deployment. If you are interested we could update to do it once Quarto pre-release is available in the Quarto plugin.

cderv:
> Currently website is still automatically deployed (but failling) at the old website

And this is now solved. Thanks @apreshill - no more failure in the checks now. 

I also activated the non-overwrite of comment @cwickham to see if you prefer that.

cderv:
Merging this - if you prefer only a single comment from the bot, we'll tweak the setting later.

PR #678, opened 2023-03-22
Move Jupyter Notebook  Embed to permanent location

cwickham:
Moved to: Guide/Authoring/Jupyter Notebooks

Not added to list on https://quarto.org/docs/guide/


jjallaire:
Location in menu LGTM! Not sure whether we would put it in the "Guide" (as only a subset show up there).

We probably also want a brief summary/callout to this in the Python article: https://quarto.org/docs/computations/python.html (maybe after "Workflow"?)

cwickham:
@jjallaire Added a link to `_jupyter-authoring-tools.md` which gets included in both the Python and Julia pages.

jjallaire (approved):

PR #679, opened 2023-03-22
Move Grid Customization Highlight Docs into Permanent Home

cwickham:
Added to Advanced/Page Layout, plus linked from a few logical places.


docs/output-formats/page-layout.qmd
```diff
@@ -2,10 +2,22 @@
 title: Page Layout
 format: html
 page-layout: full
+aliases:
+  - /docs/prerelease/1.3/grid.html
```
cwickham, line 6: @dragonstyle This isn't quite right because the old page is now only a section of this page. Is there a way to redirect from `/docs/prerelease/1.3/grid.html` to `/docs/output-formats/page-layout.qmd#grid-customization`?

jjallaire (approved):

docs/output-formats/page-layout.qmd
dragonstyle: There isn't a way to target the individual section, unfortunately.

PR #680, opened 2023-03-22
Move code annotation to Guide/Authoring/Advanced Authoring

cwickham:
Move to Guide/Authoring/Advanced Authoring

@dragonstyle and I couldn't come up with a better place. This introduces an additional section to Authoring. but this is currently the only page in it.


jjallaire:
I would have this be a top level article in "Authoring" rather than have it be under "Advanced"

We should also cross link to this from the HTML Code Blocks article (https://quarto.org/docs/output-formats/html-code.html) as well as perhaps from just below the "Syntax Highlighting" section in the other format guides (I believe that "Syntax Highlighting" is handled via an include so it should just be a matter of adding this in one place.

cwickham:
Changes made, ready for another look.

jjallaire (approved):

PR #681, opened 2023-03-22
Move Asciidoc Books Docs to Permanent Home

cwickham:
Bit of a lazy plop of all current highlight content into Guide/Books/Customizing Output

@dragonstyle It seems like "Previewing ..." is the only section with content that isn't implied by existing docs. An alternate option might be to just move that to an extra subsectiion of [Guide/Books/Creating a Book#Preview](https://quarto.org/docs/books/#book-preview). What do you think?


jjallaire:
I agree that most of the content in this new section is already covered elsewhere in the book docs so we don't need to repeat it. I feel though that if we just bolted it into the existing "Book Preview" section it might obfuscate the Publishing content below. What about if we added a new "Asiidoc Books" section below the "Publishing" section (and provide a hyperlink to it from the "Book Preview" section)

cwickham:
> What about if we added a new "Asiidoc Books" section below the "Publishing" section (and provide a hyperlink to it from the "Book Preview" section)

Sounds like a good plan. Also moved the callout about `asciidoc` vs `asciidoctor` to the reference page which like like a better home for it. @jjallaire do you want to take another look?

jjallaire (approved):

PR #688, opened 2023-03-24
Updated RStudio version and link to posit.co

NeubertJonas:
Newest RStudio version 2023.03 is now mentioned and recommended. Clarified that at least v2022.07 is required to work with some of the integrated Quarto features. Links updated to new posit.co website.


jjallaire:
Thanks!

PR #689, opened 2023-03-24
Fixed typo in Black-Scholes equation

bvancil:
"C" should be "S". See https://en.wikipedia.org/wiki/Black%E2%80%93Scholes_equation.


jjallaire:
Thanks!

PR #698, opened 2023-04-04
State unique affiliations (?) in authors.qmd

baptiste:
- trivial typo
- clarify that only unique affiliations are listed in affiliations – I assume this is the case, but it's not explicitly stated


baptiste:
It does seem to remove duplicates, as per [maybeAddAffiliation](https://github.com/quarto-dev/quarto-cli/blob/ad4281d71317c6a491434a080a408de2aff789a4/src/resources/filters/common/authors.lua#L346)

dragonstyle (approved):

PR #699, opened 2023-04-07
Update `include-in-*` documentation to be the same for PDF and HTML

cderv:
Closes quarto-dev/quarto-cli#4599

New doc

* https://deploy-preview-699--mystifying-jepsen-fa4396.netlify.app/docs/output-formats/pdf-basics.html#latex-includes
* https://deploy-preview-699--mystifying-jepsen-fa4396.netlify.app/docs/output-formats/html-basics.html#includes

Current at
* https://quarto.org/docs/output-formats/pdf-basics.html#latex-includes
* https://quarto.org/docs/output-formats/html-basics.html#includes


docs/output-formats/_document-options-includes.md
```diff
@@ -1,38 +1,9 @@
...
-For example:
-
-``` yaml
-format:
-  pdf: 
-    header-includes: |
-      \usepackage{eplain}
-      \usepackage{easy-todo}
-```
+You can specify directly a single file or multiple files for each of these options (or use the `file:` subkey), but also use the `text` sub key to include raw content directly.  Note that `|` character needs to be used on the line with `text:` inside `include-in-*:` to indicate that the value is a multi-line string. If you omit `file:` or `text:`, Quarto assumes you meant a file.
```
cwickham, line 9: Some rewording suggestions:
```suggestion
You can specify a single file or multiple files for each of these options directly, or use the `file:` subkey. To include raw content in the YAML header, use the `text` subkey.  When using `text:`, add the `|` character after `text:` to indicate that the value is a multi-line string. If you omit `file:` or `text:`, Quarto assumes you are providing a file.
```

Is "subkey" the right word for these? I think I've used suboption before but I'm not sure the correct terminology. I noticed that you used both "subkey" (no space) and "sub key" (with a space), so we should pick one and be consistent.

cwickham (changes requested):
Looks good. Just a few wording suggestions for clarity.

docs/output-formats/_document-options-includes.md
cderv: > Is "subkey" the right word for these? I think I've used suboption before but I'm not sure the correct terminology.

I don't know which is the correct one - we can choose. I often uses key and subkey in R Markdown  ecosystem because in YAML we often read key/value pairs like in the YAML spec

allenmanning (approved):
LGTM

cderv:
I merged this. There is something off with this PR preview. It works ok on other, so maybe this is because I did some modification while this PR was open. I'll check with another PR and adapt our workflow otherwise

cderv:
The merge did not solve the issue, so it hinted me it was not netlify related. In fact it is a `freeze: true` issue. This modified file here needs to be re-rendered individually to update the freeze. 

but it can even be unfrozen as **knitr** is not needed anymore. I asked question on slack, I'll update the website after . 

@cwickham We need to decide what to do with freeze. This currently require a local render to update, and as it is `freeze: true` it means a non-project render is required.

PR #703, opened 2023-04-13
Fix typo in shortcodes documentation

daniel-smit-haw:


allenmanning:
Thank you for contributing @daniel-smit-haw, once we fix our checks, we will merge in this fix.

PR #718, opened 2023-05-02
Docs for scholarly front matter

cwickham:
WIP


docs/authoring/front-matter.qmd
```diff
@@ -0,0 +1,425 @@
...
+|                      |                | and/or deceased.                                                                                                                                                    |
++----------------------+----------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| `id`                 | string         | An identifier to be used to refer to this author in other fields. See an example in [Funding].                                                                      |
++----------------------+----------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| `number`             | number         | A number to identify this author.                                                                                                                                   |
```
cwickham, line 123: @dragonstyle Would a writer ever specify `number`? Why?

dragonstyle: I think we should omit this from this documentation. We automatically add a number to the author data when we normalize it in order to permit title blocks to do things like use a superscript number to refer to the author. You're right that it isn't all that useful here!

docs/authoring/front-matter.qmd
```diff
@@ -0,0 +1,425 @@
...
+```
+
+You can read about the other keys you can provide to `author` and `affiliation` in the corresponding [Author](#author) and [Affiliation](#affiliation) sections below.
+
+Both `author` and `affiliation` can take multiple elements to describe multiple authors, or authors with multiple affiliations. As an example, here is the YAML to describe a document with two authors, the first of which has two affiliations:
+
+``` yaml
+---
+author:
+  - name: Norah Jones
```
dragonstyle, line 85: I think it might be useful to include a section (perhaps down below), on author names that illustrates providing a first name and last name. I think this will be rare, but it may be useful where the user is including an author with a complex name that we don't parse / divide well...

docs/authoring/front-matter.qmd
```diff
@@ -0,0 +1,425 @@
...
+| `country` `postal-code` |           |                                                                                                    |
++-------------------------+-----------+----------------------------------------------------------------------------------------------------+
+| `url`                   | String    | Affiliation's website. Converted to a link in many formats.                                        |
++-------------------------+-----------+----------------------------------------------------------------------------------------------------+
+| `isni`\                 | Numeric   | Affiliation's:\                                                                                    |
+| `ringgold`\             |           | [International Standard Name Identifier (ISNI)](https://isni.org/);\                               |
+| `ror`                   |           | [Ringgold ID](https://www.ringgold.com/ringgold-identifier/);\                                     |
+|                         |           | [Research Organization Registry (ROR) ID](https://ror.org/).                                       |
```
dragonstyle, line 181: ROR should be a string starting with `https://ror.org/` and followed by a 9 digit alphanumeric identifier.

docs/authoring/front-matter.qmd
```diff
@@ -0,0 +1,425 @@
...
+| `city`\                 |           |                                                                                                    |
+| `region` or `state`\    |           |                                                                                                    |
+| `country` `postal-code` |           |                                                                                                    |
++-------------------------+-----------+----------------------------------------------------------------------------------------------------+
+| `url`                   | String    | Affiliation's website. Converted to a link in many formats.                                        |
++-------------------------+-----------+----------------------------------------------------------------------------------------------------+
+| `isni`\                 | Numeric   | Affiliation's:\                                                                                    |
+| `ringgold`\             |           | [International Standard Name Identifier (ISNI)](https://isni.org/);\                               |
```
dragonstyle, line 179: isni 16 digit identifier
ringgold: 4-6 digit numeric identifier

docs/authoring/front-matter.qmd
```diff
@@ -0,0 +1,425 @@
...
+  name: Josiah Carberry
+  orcid: 0000-0002-1825-0097
+  url: https://en.wikipedia.org/wiki/Josiah_S._Carberry
+  email: josiah@psychoceramics.org
+  corresponding: true
+  affiliation: 
+    - name: Brown University
+      department: Psychoceramics
+      city: Providence
+      region: RI
```
dragonstyle, line 200: Maybe we should show state here? not sure what is better, being more international or being easier for US to understand...

docs/authoring/front-matter.qmd
```diff
@@ -0,0 +1,425 @@
...
+
+## Abstract
+
+You can add an abstract with the `abstract` key. Since abstracts are generally longer than a line, and may contain markdown, you'll need to provide it using YAML's folded block style. That is, place a `>` on the same line as `abstract:` and indent your raw abstract text by two spaces.
+
+For example:
+
+``` yaml
+---
+abstract: >
```
dragonstyle, line 273: TIL there is a different between `>` and `|` in yaml multiline strings. I had no idea and typically used pipe myself- not sure if you were aware and were suggesting `>` because in abstracts this makes sense, but wanted to raise this in case it wasn't intentional.

https://yaml-multiline.info

dragonstyle (approved):
Looks excellent and a huge improvement over the non-existent documentation! 

Should we add somewhere that this metadata will generally be supported in academic / journal templates?

docs/authoring/front-matter.qmd
cwickham: It was intentional, but now I see it's not useful given how it all gets parsed through Quarto. I'll switch to `|`.

PR #729, opened 2023-05-19
Add information about bibliography processing for typst format

cderv:
Companion PR to https://github.com/quarto-dev/quarto-cli/pull/5575


jjallaire:
We should update in our YAML schema as well if necessary.

jjallaire:
While we're at it I think we could improve the "Getting Started" directions at the top by providing instructions/link for installing the v1.4 pre-release

PR #741, opened 2023-05-26
Add clean reveal.js theme

grantmcdermott:
Live demo here: https://grantmcdermott.com/quarto-revealjs-clean-demo/template.html

### DIFF
--- a/docs/extensions/listings/revealjs-formats.yml
+++ b/docs/extensions/listings/revealjs-formats.yml
@@ -52,3 +52,9 @@
   author: "[James Joseph Balamuta](https://github.com/coatless)"
   description: |
     University of Illinois Urbana-Champaign inspired theme for the reveal.js format based on Metropolis.
+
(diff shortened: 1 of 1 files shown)


grantmcdermott:
P.S. Should these entries be ordered alphabetically? I'm happy to do so if you think that is less arbitrary than the current chronological ordering...

jjallaire:
They are somewhat arbitrary. Ordering by likely user interest would be an aspiration for sure!

cderv:
> Ordering by likely user interest would be an aspiration for sure!

We could imagine something based on some repo stats (like number of Stars ?)

* Query the repo source URL information
* Order based on the stat chosen 
* Output the content in the new order inside quarto doc
* To not query all the time we could have some caching in a file with month and year, and only update every new month.

Probably requires some adjustment but using R chunks, there is all the tools for this IMO

I believe something like this could work - Just ideas.

PR #747, opened 2023-05-31
Add  a callout note about `code-link` not working when used with `code-line-numbers`

cderv:
Related to https://github.com/quarto-dev/quarto-cli/issues/2036

Adding a note about this known limitation seems important for users to easily know about it. 

@cwickham I used a callout note here, but maybe you will think that a callout warning is better. 

https://deploy-preview-747--mystifying-jepsen-fa4396.netlify.app/docs/output-formats/html-code.html#code-linking


mcanouil:
FWIW, I am concerned about cluttering the documentation with notes about opened issues more or less easy to discover on GitHub.
Which issues get a note in the documentation and which don't?

cderv:
> FWIW, I am concerned about cluttering the documentation with notes about opened issues more or less easy to discover on GitHub.

Do you think this one is cluttering the doc ? 

> Which issues get a note in the documentation and which don't?

This is a limitation of a Quarto feature that is no documented. This callout aims at documenting it. Linking to the issue about that is just for transparency, and for users to be able to follow possible improvement. All the more because it is not a Quarto issue, but a limitation about the tooling that Quarto uses to bring the feature (here **downlit**). 

We could also just documented without callout and without reference to any other issue that you can't use `code-link: true` with `code-line-numbers: true`. I am fine by adding a more affirmative sentence with no details on why. 

To me this is not about documenting an issue, it is about documenting a limitation or rather detailing the usage context of a feature. 

That is how I think this is worth adding.

mcanouil:
Also, I believe there is a need for a PR in Quarto CLI for the reference pages.
Otherwise, it's unlikely the information will be find easily.

cwickham:
Yes, that's what I mean. 

I think the [second sentence in the schema]( https://github.com/quarto-dev/quarto-cli/blob/637636365a600fd6cc79bae685c9853f5ffc4f6b/src/resources/schema/document-code.yml#L28-L33) could say:

> Code linking is currently implemented only for the knitr engine (via the [downlit](https://downlit.r-lib.org/) package). A limitation of downlit currently prevents code linking if `code-line-numbers` is also `true`.

cderv:
@cwickham I used the sentence in both places. Companion PR is https://github.com/quarto-dev/quarto-cli/pull/5768

cwickham (approved):
Added the updated reference page from running `tools/reference.ts`. Otherwise, looks great.

PR #748, opened 2023-05-31
Note requirement for cell options

edavidaja:
see https://github.com/quarto-dev/quarto-cli/issues/5767


cwickham:
Thanks @edavidaja! Text change looks good, but I think this whole paragraph shouldn't be buried in the Graphviz section. I added a commit that moved things around a bit.

cwickham (approved):

PR #751, opened 2023-06-02
Corrected text Update _computations-complete.qmd

Janidai:
The plots in @fig-mpg show the relationship between city and highway mileage for 38 popular models of cars.
In @fig-mpg-1 the points are colored by the number of cylinders while in @fig-mpg-2 the points are colored by engine displacement.


cwickham:
Thanks for the PR!

PR #754, opened 2023-06-08
Update carousel.yml

jfunction:
While checking out the site I noticed the carousel tile for interactive documents was targeting a non-existent id. I just renamed it to be in line with https://github.com/quarto-dev/quarto-web/blob/main/docs/gallery/index.qmd#L29 which works.

### DIFF
--- a/docs/gallery/carousel.yml
+++ b/docs/gallery/carousel.yml
@@ -9,7 +9,7 @@
   
 - caption: "Engage readers with interactivity"
   image: "interactive/interactive-jupyter.png"
-  link: "#interactive-documents"
+  link: "#interactive-docs"
 
 - caption: "Publish collections of articles as a website"
   image: "websites/websites-quarto.png"


jfunction:
Starting to think the bot deployed the wrong version. The carousel is implemented by [pulling in carousel.yaml](https://github.com/quarto-dev/quarto-web/blob/main/docs/gallery/index.qmd#L38) which I [changed](https://github.com/quarto-dev/quarto-web/pull/754/commits/ab5ffb4190a896e5f7a8224ff6dca71130fedde4).

jfunction:
Ah - of course - I need to update https://github.com/quarto-dev/quarto-web/blob/main/_freeze/docs/gallery/index/execute-results/html.json
I will do that on Monday.

cderv:
Oh yes, if the document is frozen a special render of the document needs to be done. The preview only run `quarto render` on the whole project and this website has `freeze: true`. So manul update of the frozen state needs to be done before merging.

jfunction:
Still keen to finish this - I could probably just manually edit the _freeze files but would rather to do this the right way. Just waiting on https://github.com/rstudio/renv/issues/1343

jfunction:
Looks good to me now:
If you go to https://deploy-preview-754--mystifying-jepsen-fa4396.netlify.app/docs/gallery/
Then click any image on the carousel, it should take you to the appropriate section of the page.

cderv (approved):
Thank you!

jfunction:
btw I'm considering writing an extension so the carousel renders in pandoc via Lua which will be a bit cleaner. If I get around to it I will have a go at implementing that extension here also (though there should be no change to the user).

cderv:
> If I get around to it I will have a go at implementing that extension here also (though there should be no change to the user).

IMO No need to use that extension here as this is working. But we could consider it if we extend the gallery to more contents !

jfunction:
Cool - thanks for your help on this :)

PR #760, opened 2023-06-18
Propose Adding docx-horizontal-rule extension

ttalVlatt:
I made a simple extensions that allows users to edit the horizontal rule output for .docx files, since that wasn't editable (as far as I could work out) in a reference-doc. Thought I would share in case you think it's useful for more users

### DIFF
--- a/docs/extensions/listings/shortcodes-and-filters.yml
+++ b/docs/extensions/listings/shortcodes-and-filters.yml
@@ -271,3 +271,9 @@
   description: |
       A Quarto filter to run SQL queries interactively from a SQLite database created at render time 
       in html documents and Revealjs presentations.
+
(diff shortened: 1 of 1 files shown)


jjallaire:
The fact using this requires editing of the .lua script makes it perhaps not ideally suited to being an extension (as extensions typically provide configuration/options to tailor their behavior). Perhaps you could add something like `docx-hr-thinkness` and `docx-hr-color` options that you could read within the extension?

ttalVlatt:
Hi there, I've gone ahead and updated to extension to now take `docx-hr-x` options from the YAML to edit the line output. Let me know if there's anything else you'd like to see added/changed!

jjallaire:
Awesome, that looks great!

PR #761, opened 2023-06-18
fix: github action checkout v2 deprecated

mcanouil:
GitHub is entering its final deprecation stage of Node12 which is used in the checkout v2 action.
See <https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/>.

Actions in <https://github.com/quarto-dev/quarto-actions> are already using V3, but the website was still showing v2.

This PR fixes the actions, in the website itself and in the CI of this repository.


cderv (approved):
Thanks

PR #769, opened 2023-06-27
tweak overview language

jjallaire:
Here's a proposed tweak to the overview language. The main thing I did was:

(1) Try to distinguish manuscripts from normal notebooks or qmds by mentioning that you can use "one or more" of them (to me just saying "Link your computations to your narrative, allowing your readers to dive into your code" isn't quite enough to see why this is something new).

(2) Punch up the emphasis of "multiple formats" available through the website (again to distinguish from a normal article you might publish)

(3) Eliminate the reference to MECA bundles in the intro (too much to try to communicate/understand, and in the early days this will only be interesting to users that have been pointed to us from a journal that is prepared to accept MECA submissions.

### DIFF
--- a/docs/manuscripts/index.qmd
+++ b/docs/manuscripts/index.qmd
@@ -16,13 +16,13 @@ format:
 
 ## Overview
 
-Quarto provides a framework for writing and publishing your scholarly articles. A Quarto manuscript lets you:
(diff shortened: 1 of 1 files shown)


cwickham:
Looks great, thanks!

PR #772, opened 2023-06-28
Properly place guide and reference in sections in sidebar

dragonstyle:


cwickham:
What do you think about adding the same treatment for "Extensions"? On one hand, Extensions feels like it plays the same kind of role as Guide and Reference. On the other hand, the subsections of Extensions include Extension in the name ("Quarto Extensions", "Creating Extensions"), so perhaps it is more obvious where you are.

As a side note, this makes me think we need a page somewhere (maybe the Guide) that describes "Using Extensions".

dragonstyle:
> What do you think about adding the same treatment for "Extensions"?

I think you're right about needing to add something more about actually using extensions. We currently are sort of conflating:
- extension discovery (e.g. the gallery of extensions)
- how to use extensions (its actually sort of buried under 'managing extensions')
- how to develop extensions (most of the tree)

I tried tinkering with different structures with 'Extensions' as a top level page, but couldn't really find anything that I thought worked. 

I do wonder whether we should promote 'Managing Extensions' to a more discoverable position (perhaps as a peer to the others) and renaming it to something like 'Using Extensions'

dragonstyle:
> I do wonder whether we should promote 'Managing Extensions' to a more discoverable position (perhaps as a peer to the others) and renaming it to something like 'Using Extensions'

idk if that really works...

Screenshot 2023-06-29 at 3 56 53 PM

I'm going to go ahead and merge this for now since it an improvement and we can keep working on figuring out 'Extensions' if you're ok with that.

PR #776, opened 2023-06-30
Front matter fixes

cwickham:
Fixes problems reported under Remarks in: https://github.com/quarto-dev/quarto-cli/discussions/5860#discussioncomment-6290755

### DIFF
--- a/docs/journals/authors.qmd
+++ b/docs/journals/authors.qmd
@@ -4,7 +4,6 @@ title: "Authors & Affiliations"
 
 {{< include ../extensions/_extension-version.qmd >}}
 
-
 ## Overview
 
 An important goal for Quarto is to make it possible to use the same source document to produce multiple output formats. One significant challenge to overcome is defining a consistent way to express author and affiliation metadata such that articles targeting multiple Journals do not require special tweaking of authors and affiliations for each publication.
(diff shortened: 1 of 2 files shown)


docs/journals/authors.qmd
```diff
@@ -153,6 +155,37 @@ author:
...
+-   An array of key-value pairs of the form `role: contribution`:
+
+    ``` yaml
+    author: 
+      role: 
+        - conceptualization: lead
+        - methodology: supporting
+    ```
+
+If a role matches one of the [CRediT roles or their aliases](https://github.com/quarto-dev/quarto-cli/blob/f65180a75e1cf2996328cd51cb4fd5d02d391511/src/resources/filters/modules/authors.lua#L119-L144), vocabulary decorations are added.
```
cwickham, line 186: @dragonstyle I just pulled language from the code. How would you describe what is happening here to a template author?

dragonstyle: What about something like:

```markdown
If a role matches one of the [CRediT roles or their aliases](https://github.com/quarto-dev/quarto-cli/blob/f65180a75e1cf2996328cd51cb4fd5d02d391511/src/resources/filters/modules/authors.lua#L119-L144), the additional properties `vocab-identifier`, `vocab-term`, and `vocab-term-indentifier`, will be added to the role with the appropriate value from the CRediT specification.
```

It's more or less what you already have...

PR #778, opened 2023-07-09
Update html-themes.qmd

jdonland:
Correct erroneous note on $footer-bg variable.


jjallaire:
Thanks for this! Grid tables need their columns to line up exactly --- could you tweak the column indicator that is now offset 1 to the right?

cderv (approved):
Thanks

PR #781, opened 2023-07-12
Add examples of using divs to separate content

cwickham:
Closes https://github.com/quarto-dev/quarto-cli/issues/4600

Preview Changes:
* [Lists](https://deploy-preview-781--mystifying-jepsen-fa4396.netlify.app/docs/authoring/markdown-basics.html#lists)
* [Block Layout](https://deploy-preview-781--mystifying-jepsen-fa4396.netlify.app/docs/authoring/figures.html#block-layout)

### DIFF
--- a/docs/authoring/figures.qmd
+++ b/docs/authoring/figures.qmd
@@ -518,3 +518,26 @@ Note that headings are automatically combined with the block that follows them, 
 Lorem ipsum dolor sit amet, consectetur adipiscing elit. Curabitur gravida eu erat et fring. Morbi congue augue vel eros ullamcorper, eget convallis tortor sagittis. Fusce sodales viverra mauris a fringilla. Donec feugiat, justo eu blandit placerat, enim dui volutpat turpis, eu dictum lectus urna eu urna. Mauris sed massa ornare, interdum ipsum a, semper massa. 
 :::
 ```
+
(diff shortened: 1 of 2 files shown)


mcanouil:
For the "list" part the output seems off as the code should produce compact list but we can see a spaced list on the right.
image

The "block layout" is great.
Now the tricky question is how to make it discoverable in the search as most users will likely search for "Multiple Columns" which leads only to `:::: {.columns}` solution.

mcanouil:
I would suggest to refactor the markdown basic document to not use the `.    ` code block syntax and use `` ```{.markdown} `` on the left column, and use the exact same markdown outside the code block in the right column.

Advantages:
- output will actually be the one from the code
- code will have the copy button

cwickham:
I agree with the suggestion to use ` ```{.markdown}`. Interestingly, the visual editor is converting it to ` ``` ` automatically for me. Do you see any reason not to use ` ```markdown`, over ` ```{.markdown} `? ` ```markdown` would involve a lot less fighting with the visual editor on my end.

mcanouil:
Actually it's the same as long as it is a code block with the right language.

cwickham:
Thanks for the suggestions @mcanouil. Refactored this page to use ` ```markdown` and made sure the list code matches the list output. I think we can address discoverability with https://github.com/quarto-dev/quarto-cli/issues/6192.

PR #788, opened 2023-07-18
Fix link to "manuscript basics"

mine-cetinkaya-rundel:
Fix link to "manuscript basics" (file is called components.qmd)


jjallaire:
Since this is a grid table, you need to make sure that the column `|` is aligned after the edit (I see its shifted over a few characters in this PR). Safest way I have found for making tweaks to grid tables is to just use the visual editor.

PR #792, opened 2023-07-20
fix: extension embedding example

mcanouil:
This PRs fixes and improves the documentation for extension embedding.

The example was to embed extension in an installed extension, which actually should never happen in user, but instead you want to embed extension when developping one.

In addition, the `jss` / `fancy-text` example now got adirectory tree to clearly show that embedding means adding an extension within an extension.

This is related to quarto-dev/quarto-cli#6291

### DIFF
--- a/docs/extensions/_formats-common.qmd
+++ b/docs/extensions/_formats-common.qmd
@@ -58,6 +58,7 @@ In some cases format extensions will want to make use of other extensions. This 
 
 ``` {.bash filename="Terminal"}
 quarto create extension format:pdf myformat
+cd myformat
 quarto add quarto-ext/fancy-text --embed myformat
 ```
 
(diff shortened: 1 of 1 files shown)


docs/extensions/_formats-common.qmd
```diff
@@ -56,12 +56,29 @@ contributes:
 
 In some cases format extensions will want to make use of other extensions. This is permitted, but adding extensions for use within a custom format must be done with a special command line flag to ensure they are embedded correctly.
 
+``` {.bash filename="Terminal"}
+quarto create extension format:pdf myformat
+quarto add quarto-ext/fancy-text --embed myformat
```
cderv, line 61: This is missing a step to be working by a copy paste
```suggestion
quarto create extension format:pdf myformat
cd myformat
quarto add quarto-ext/fancy-text --embed myformat
```

Also this is causing Deno panick on Windows but I am looking into that.

cderv: The deno panic issue is now found and solve. Nice finding ! :)

PR #795, opened 2023-07-24
Add a landing page for Reference > Code Cells

cwickham:
I'd like to provide a link on the cheatsheet for where people can learn about other code cell options. Currently, I'd need to list both a knitr and jupyter link because there is no common landing page (except the top-level Reference page).

I'm not sure if there is a less duplicative approach than what I did here (copying part of the top-level `reference.yml`)...

Also open to what text might be on this landing page.


jjallaire:
I think this is a fine approach (and agree its great to have a single target).

PR #799, opened 2023-08-01
Update rstudio.qmd

Sumidu:
added a note for inline R Code evaluation.

The note explains the changes that require an explicit rendering engine definition if no code chunks are available.


cwickham:
Thanks for this @Sumidu! I think this is an important caveat to point out, but perhaps this particular page isn't the right spot. Given this is the "Getting Started" section, I think most readers will be working with the template files which do have R code chunks in them. 

Would you be willing to propose a change to this section instead?https://quarto.org/docs/computations/execution-options.html#inline-code 

I think the best format might be as a [callout](https://quarto.org/docs/authoring/callouts.html).

cwickham:
Closing this, as now captured in #800

PR #800, opened 2023-08-03
Added a note about engine specification for inline R code

Sumidu:
Since 1.3.433 it is necessary to specify the engine (knitr) if no code chunk uses R. 
This was not documented anywhere else. So I added a callout-note in lines 181-187.

The second amendment in the file was created by GitHub. I did not actively edit line 415. Maybe the GitHubs Code editor added a CR-LF at the end?
I hope this is not a problem


cwickham:
Thanks @Sumidu. I closing this because it's now covered in: https://quarto.org/docs/computations/inline-code.html#engine-binding

PR #803, opened 2023-08-07
fix: add a note to install Jupyter extension

mcanouil:
Fixes quarto-dev/quarto-cli#6437


jjallaire:
I think the right fix is to make sure these are always installed. I did that here: https://github.com/quarto-dev/quarto/commit/7a94f8c769091bceb1853a1ffa2c0c61b3625a1f

mcanouil:
Makes sense and definitely better.

PR #814, opened 2023-08-30
Add missing column specifiers

jhchou:


cwickham:
Fixed in https://github.com/quarto-dev/quarto-web/pull/957

PR #816, opened 2023-08-31
Add example website to gallery

dkapitan:
I would love to showcase the new website that I build using Quarto in the gallery.


dkapitan:
Thanks, looks good to me.

PR #833, opened 2023-09-21
split up revealjs sass variables table into multiple

EmilHvitfeldt:
I enjoyed the way the tables were split into categories on the [html sass variables](https://quarto.org/docs/output-formats/html-themes.html#sass-variables), and thought it would be neat if that could be replicated on the revealjs side


jjallaire:
Thanks!

PR #846, opened 2023-10-04
devito website example

mloubout:
New website example that is also built in combination with [quartodoc](https://github.com/machow/quartodoc) for autogeneration of the API reference.


mloubout:
> You might want to take a look at your HTTPS certificate, I think there's a misconfiguration there.

Did not see anything wrong but will double check, thanks for the heads up.

cscheid:
> Did not see anything wrong but will double check, thanks for the heads up.

Huh. On reload I can't see it, but my first attempt said something about the certificate pointing to github pages instead of devitoproject.org. Everything looks good with the letsencrypt cert I see now. I wish I had taken a screenshot; sorry I can't reproduce it anymore!

dragonstyle:
I ran into the same issue:

Screenshot 2023-10-04 at 12 07 51 PM

mloubout:
Looks like something went off in the migration from the previous website, will fix it asap, thanks so much for bringing it up.

PR #852, opened 2023-10-05
Change "Knitr" to "knitr" in the index.qmd file

chuxinyuan:
yihui uses knitr at all times, so I think the original author should be respected.


cderv (approved):
Thanks

PR #877, opened 2023-10-18
Add  quarto-ieee template for IEEE transaction/journal

dfolio:
IEEE is one of the most famous scientific associations in the fields of engineering. It is the publisher of hundreds of peer-reviewed transactions, journals and magazines in various fields (e.g. electronics, computer engineering, control, biomedical, etc.).
This adds the [quarto-ieee] template, which supports the [`IEEEtran`](https://ctan.org/pkg/ieeetran)  LaTeX document class for PDF output. In addition, the HTML output is supported, and attempts to mimic as closely as possible the paper layout seen on [IEEEXplore<sup>®</sup>](https://ieeexplore.ieee.org/).

The [quarto-ieee] template is hosted on Github, includes a [`README.md`](https://github.com/dfolio/quarto-ieee#readme), and is licensed under the MIT.

[quarto-ieee]: <https://github.com/dfolio/quarto-ieee>

### DIFF
--- a/docs/extensions/listings/journal-articles.yml
+++ b/docs/extensions/listings/journal-articles.yml
@@ -81,3 +81,10 @@
   author: "[christopherkenny](https://github.com/christopherkenny)"
   description: |
     American Political Science Review (APSR)
+    
+- name: ieee
+  path: https://github.com/dfolio/quarto-ieee
+  author: "[dfolio](https://github.com/dfolio)"
+  description: |
+    Institute of Electrical and Electronics Engineers (IEEE)
+


cderv (approved):
Thank you

cderv:
It seems something is not working with 1.4 though to render your template. It could be an issue on our side. 

You should probably test your template against latest pre-release to find issue early, and report to us if this is a bug on our side (and not a change in quarto that requires adaptation in Lua for example)

dfolio:
It strange it works properly with 1.3.450.

I have checked with 1.4.429, and I effectively get the error

```bash
ERROR: 
compilation failed- missing packages (automatic installed disabled)
Undefined control sequence.
l.307 \citeproc
```

Is it the same error that you saw from your side?

A first fix is to use `cite-method: biblatex` instead of `citeproc`.
But I have just seen that the `default.latex` pandoc template now defines a `citeproc` command that was missing from my previous `ieee-template.tex`.
I have corrected/checked, and it works from my end.
Let me know if it works for you.

I think that such error may affect many users that use their own template instead of the default one with `citeproc`.

cderv:
> compilation failed- missing packages (automatic installed disabled)

Are you using TeX Live ? Automatic installation would be disabled if not using TeX Live as LaTeX distribution, or explicitly opt-out. 
Quarto uses by default a Quarto engine that does try to install missing dependencies (https://quarto.org/docs/output-formats/pdf-engine.html#quarto-pdf-engine)

> A first fix is to use cite-method: biblatex instead of citeproc.
> But I have just seen that the default.latex pandoc template now defines a citeproc command that was missing from my previous > ieee-template.tex.

By default, Quarto will default like Pandoc to using Pandoc's citeproc for citation processing. but you can easily change to another `cite-method` by default for your format. This is detailed in there: https://quarto.org/docs/output-formats/pdf-basics.html#citations
For Pandoc's citeproc to work, you need indeed to be sure that the necessary thing for the Pandoc's LaTeX template are included. To ease this, Quarto has introduced the logic of Partials for the template. The one you need is [citation.tex()](https://github.com/quarto-dev/quarto-cli/blob/main/src/resources/formats/pdf/pandoc/citations.tex). See about partials there: https://quarto.org/docs/journals/templates.html#latex-partials
You should be able to leverage existing partials, even if you provide a custom template. 

Happy to discuss this further in an issue in your repo. Do not hesitate to ping me there.

dfolio:
Thx for your appreciation

Effectively, I use TexLive.
I found two necessary packages to be installed from tinytex (physics and orcid).
I would like to know, what is the best (recommended) practice:
- set `latex-auto-install: true` 
- put the missing latex package as  `format-resources:`

I'm already using Partials and will try to improve by taking advantage of existing Partials (e.g. using `$pandoc.tex()$`).

cderv:
> Effectively, I use TexLive.

Then automatic installation should not be disabled... 🤔 
Do you have writing rights on your TeX Live installation ? Or maye it is from a system installation ? Did you use [`quarto install tinytex`](https://quarto.org/docs/output-formats/pdf-engine.html#installing-tex) ? 

> set latex-auto-install: true

This should be the default as explained in the doc with Quarto engines, unless you deactivate it. 

> put the missing latex package as format-resources:

This won't work I think. Unless you provide a copy indeed. But with auto install, the missing packages should be detected and automically installed. It seems something prevent this on your installation. 

What is you `quarto check install` (if you are using 1.4 version) ?

dfolio:
> Then automatic installation should not be disabled... 🤔

When I was using TexLive I didn't have automatic installation, nor when I was using  [`quarto install tinytex`](https://quarto.org/docs/output-formats/pdf-engine.html#installing-tex)  (`latex-auto-install: true` is necessary to update tinyTeX). However, with TexLive, all requested LaTeX's packages are already available.

>This should be the default as explained in the doc with Quarto engines, unless you deactivate it.

I don't have modified any default behavior.
Note that I'm using a Gentoo distribution that doesn't use `tlmgr` natively. I have to perform a manual installation (as in [https://wiki.gentoo.org/wiki/TeX_Live_manual_installation](https://wiki.gentoo.org/wiki/TeX_Live_manual_installation)) to get it.
After the manual installation (and removing all Gentoo TexLive packages), the automatic installation work properly  (`latex-auto-install: true` is not necessary). Therefore, I can assume the issue is mainly related to the gentoo distribution.

> What is you `quarto check install` (if you are using 1.4 version) ?

`quarto check install` output

```bash
Quarto 1.4.429
[✓] Checking Quarto installation......OK
      Version: 1.4.429
      Path: /opt/quarto-1.4.429/bin

[✓] Checking tools....................OK
      TinyTeX: v2023.10
      Chromium: (not installed)

[✓] Checking LaTeX....................OK
      Using: TinyTex
      Path: /home/dfolio/.TinyTeX/bin/x86_64-linux
      Version: 2023

[✓] Checking basic markdown render....OK
```

Can you check if `quarto-ieee` works properly from your side? Thx

cderv:
> Can you check if quarto-ieee works properly from your side? Thx

Using the provided template ?

PR #888, opened 2023-10-22
Code listings: render the callout

kcarnold:
The callout-note was showing up as part of the code example, which I don't think was what was intended.

### DIFF
--- a/docs/authoring/cross-references.qmd
+++ b/docs/authoring/cross-references.qmd
@@ -267,6 +267,7 @@ SELECT * FROM Customers
 ```
 
 Then we query the customers database (@lst-customers).
+````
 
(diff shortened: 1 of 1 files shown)


cscheid:
That's correct, thank you.

In case you found it because you were looking for this feature, do note the callout is now outdated, because there is a way to get filenames on executable listings on the 1.4 prerelease:

<table>
<tr><td>

````
---
title: executable code listings
---

```{r}
#| label: fig-1
#| fig-cap: A test
#| lst-label: lst-1
#| lst-cap: The listing
plot(1:10)
```

See @fig-1 and @lst-1.

````

</td><td>
image

</td></tr></table>

PR #889, opened 2023-10-24
Gapminder dashboard wider image

mine-cetinkaya-rundel:
The screenshot of gapminder dashboard at https://quarto.org/docs/dashboards/#walkthrough has axis labels that are overlapping. This one is a wider image where they don't overlap. I'm not sure if it's necessarily better.


jjallaire:
I think yours is better but maybe a bit too wide? (I don't super-mind the overlapping axis labels, at least not enough to distort the natural orientation). What about something in between?

mine-cetinkaya-rundel:
Yeah I don't think it gets much better, I tried a few things but no real luck: https://deploy-preview-889--mystifying-jepsen-fa4396.netlify.app/docs/dashboards/#walkthrough.

I do think linking to a published version of the dashboard could be nice for experiencing interactivity and then things would look nice too, maybe that's the solution here (and then the screenshot can be left as is).

jjallaire:
The super wide is kind of a bummer (esp. on the homepage / where first impressions are made). The fact that it is viewed in shrunken form also helps. Agreed that we should just publish it as an example then we don't need to sweat it. I'll do that!

PR #892, opened 2023-10-25
add restaurant inspection example dashboard

isabelizimm:
If you're looking for more Python dashboards, here is one I created for a model on restaurant health inspections in the city of Chicago!


jjallaire:
Thank you! I am going to work on PR to make some small refinements, will send that your way soon....

PR #893, opened 2023-10-27
Adds stock explorer quarto dashboard

garrettgman:


cderv:
There is conflict in thumbnail name. 

It seems there is already an example for this added after your PR, but not same location. Should it be updated instead ? 
https://github.com/quarto-dev/quarto-web/blob/e9e585c5ac7ce1d954ba628656b098706aa4dbe8/docs/dashboards/examples/examples.yml#L3-L7

jjallaire:
We can close this (as we already deployed a fork of it)

PR #900, opened 2023-10-31
Add dashboard

ivelasq:
Hello! In case you would like another example, please find a non-Shiny Python dashboard here. I used the column orientation and tabsets to show some of the features.

So exciting about Quarto Dashboards! 🥳


jjallaire:
Thank you!

PR #904, opened 2023-11-01
add example entry for runpacer dashboard (app)

chendaniely:
YAML entry and screenshot of runpacer quarto-dashboard (it's more an app than a dashboard)


jjallaire:
Thanks! Let's hold off on merging this right now as I think I'd like it to get a bit more visual polish before we feature it. Also noted that the first page is very slow to load -- could this be a Colorado thing? (i.e. would we be better off deploying on shiny apps.io?)

PR #905, opened 2023-11-01
Expanded Shiny tutorial

jcheng5:


cderv:
@jcheng5 I saw you tried to deploy. I think it did not work because you need to be a COLLABORATOR on the project.
I added you on this project so next time it will works. You can also to branch on the repo now, which will trigger the preview without comment. 

If you don't need write access, I can reduce the permissions if this seems better. 

Hope it helps

PR #913, opened 2023-11-07
Added tabset explaining creating subtables in a python computation block

jgostick:
The R instructions call `kable` *twice* to produce the 2 tables, but in python calling `Markdown` twice does not work. Instead I found that `Markdown(table1 + table2)` seems to work, so I created a `tabset` with instructions for R and Python.

### DIFF
--- a/docs/authoring/tables.qmd
+++ b/docs/authoring/tables.qmd
@@ -279,7 +279,10 @@ kable(head(cars))
 
 If your code cell produces multiple tables, you can also specify subcaptions and layout using cell options:
 
-```{{r}}
+::: {.panel-tabset}
+## R
+
+``` {.r}
 #| label: tbl-example
 #| tbl-cap: "Example"
 #| tbl-subcap: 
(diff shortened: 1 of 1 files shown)


jjallaire:
Thanks! Another way to go for Python is to use the `display` function (e.g. `display(Markdown(cars))`). I also think we might need a `to_markdown()` call on the data frames. I'll update the example and ping this thread back when that's done.

jgostick:
Yes, your suggestion of `display(Markdown(tbl1))` is *much* preferred since mine breaks if the tables don't happen to have `\n` before and after them!  So sorry that I didn't do my due diligence on this. I was using cut/pasted tables in my IDE, and noticed just now that my real "generated" tables don't quite work the same.  Should I do another PR to fix this up?  I don't want to create a lot of noise for you over the sake of 1 line.

jjallaire:
I created a revision version here: https://quarto.org/docs/authoring/tables.html#computations

This uses `display()` and also documents how to import it as well as the use of the `index = False` parameter.

jgostick:
You are a champion, and quarto is the greatest piece of software I have used in a long, long time.  I am writing a textbook with this, so hopefully I can add it to the prestigious example gallery one day in the not too distant future.  Cheers!

jjallaire:
Thanks for the kind words and so happy to hear you are having success using Quarto for your book!

PR #917, opened 2023-11-13
remove unnecessary div

ratnanil:
This wrapper was added by mistake (?) and is confusing to the reader


cwickham:
The div here is intentional - this div appears in the intermediate markdown and prevents the code output from being interpreted as markdown (hence the need for `asis`).

I do agree that this is not particularly clear from the text. I'm going to suggest we change the sentence before the code block to: 

> If we had not specified `output: asis` the output, as seen in the intermediate markdown, would have included Quarto's `.cell-output` div:

cwickham:
Closed in favor of #920

PR #920, opened 2023-11-15
Clarify div is intentional

cwickham:
In response to https://github.com/quarto-dev/quarto-web/pull/917


ratnanil:
ok it makes more sense now. But its still a bit confusing, since `.cell-output` / `.cell-output-stdout` is not mentioned[ anywhere else ](https://github.com/search?q=repo%3Aquarto-dev%2Fquarto-web+.cell-output&type=code)in the docs.

PR #922, opened 2023-11-16
Update citation.qmd

mfenner:
CSL generates JSON (or YAML) using an XML spec. The end user never deals with the XML, so I would leave that out.


cwickham:
I think that statement clarifies that the provided link describes an XML schema but Quarto uses a YAML one. I'm going to close for now.

@mfenner thanks for raising this, it did highlight another issue with this page: https://github.com/quarto-dev/quarto-cli/issues/8713

PR #924, opened 2023-11-17
Emphasize extension options can be set in _quarto.yml

coatless:
Modifies the existing _quarto.yml to showcase how to set a global extension value for `lightbox`.

Follow up to the discussion in: 

https://github.com/quarto-dev/quarto-cli/discussions/6968

### DIFF
--- a/docs/projects/quarto-projects.qmd
+++ b/docs/projects/quarto-projects.qmd
@@ -9,7 +9,7 @@ Quarto projects are directories that provide:
 
 -   A way to render all or some of the files in a directory with a single command (e.g. `quarto render myproject`).
 
--   A way to share YAML configuration across multiple documents.
+-   A way to share YAML configuration across multiple documents, including extension options.
 
 -   The ability to redirect output artifacts to another directory.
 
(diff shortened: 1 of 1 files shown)


docs/projects/quarto-projects.qmd
```diff
@@ -70,9 +70,14 @@ format:
     documentclass: report
     margin-left: 30mm
     margin-right: 30mm
+
```
jjallaire, line 73: We might not want to use `lightbox` as the example here as this is now built in to Quarto v1.4 (so in the future not likely to be recognized as an extension).

coatless: I suppose the next option would be [latex-environment](https://github.com/quarto-ext/latex-environment); however, this lacks a single top-level key (e.g. values are set in `environments` and/or `commands`). What it be okay if I used a made-up extension? 

I avoided the other extensions in [quarto-ext/](https://github.com/quarto-ext/) since they were either: shortcodes with no global options ([fancy-text](https://github.com/quarto-ext/fancy-text), [holder](https://github.com/quarto-ext/holder), and [fontawesome](https://github.com/quarto-ext/fontawesome)), format-specific keys ([attribution](https://github.com/quarto-ext/attribution) and [pointer](https://github.com/quarto-ext/pointer)), or only had code cell level options to set ([shinylive](https://github.com/quarto-ext/shinylive) and [include-code-files](https://github.com/quarto-ext/include-code-files)).

jjallaire: I do think it is supposed to be generally understood that _anything_ you can put in document yaml can be put in `_quarto.yml`. Therefore it might be more than good enough to just comment on the fact that this also applies to extensions (rather than use an example).

cwickham:
@coatless you haven't touched this PR in a while, are you still willing to edit it to focus on emphasizing that extension YAML can also be moved to `_quarto.yml` without including a specific example?

coatless:
@cwickham I think when I was trying different combinations I discovered a merge bug between `_metadata.yml` + `_quarto.yml` but ran out of time to raise an issue in `quarto-dev/quarto-cli`.

PR #957, opened 2023-12-21
Add missing column specifiers to table

salim-b:
`page-inset-left` and `page-inset-right` were missing.

I guess the table should ideally be constructed from the [YAML schema](https://github.com/quarto-dev/quarto-cli/blob/b6f40394f4564e963bd7645eb55bb06c22fde562/src/resources/schema/definitions.yml#L58-L79), i.e. use it as the single source of truth.

A related question regarding the YAML schema: It contains the additional column specifier **`screen-rightcolumn`**, which is nowhere documented and seems odd (but is also present in [`schema-types.ts`](https://github.com/quarto-dev/quarto-cli/blob/b6f40394f4564e963bd7645eb55bb06c22fde562/src/resources/types/schema-types.ts#L52)). Is this a mistake? Quarto doesn't complain when using it (since the validation is based on this very schema file, I guess) but it doesn't seem to differ from the `body` column specifier.


cwickham (approved):
Change looks good to me. I just needed to update the freeze folder since this page has some executed code cells.

cwickham:
Thanks for the fix @salim-b! I'll merge shortly.

Let's see if @dragonstyle has any insight on this question:

> A related question regarding the YAML schema: It contains the additional column specifier screen-rightcolumn, which is nowhere documented and seems odd (but is also present in [schema-types.ts](https://github.com/quarto-dev/quarto-cli/blob/b6f40394f4564e963bd7645eb55bb06c22fde562/src/resources/types/schema-types.ts#L52)). Is this a mistake? Quarto doesn't complain when using it (since the validation is based on this very schema file, I guess) but it doesn't seem to differ from the body column specifier.

cwickham:
I'm guessing it's a typo since there is no mention of it in [_bootstrap-rules.scss](https://github.com/quarto-dev/quarto-cli/blob/1fbbc38d211dc23a51f6ed47334063cd56f214ef/src/resources/formats/html/bootstrap/_bootstrap-rules.scss#L440-L466) or [format-html-bootstrap.ts](https://github.com/quarto-dev/quarto-cli/blob/1fbbc38d211dc23a51f6ed47334063cd56f214ef/src/format/html/format-html-bootstrap.ts#L1923-L1940)

salim-b:
> I'm guessing it's a typo since there is no mention of it in [_bootstrap-rules.scss](https://github.com/quarto-dev/quarto-cli/blob/1fbbc38d211dc23a51f6ed47334063cd56f214ef/src/resources/formats/html/bootstrap/_bootstrap-rules.scss#L440-L466) or [format-html-bootstrap.ts](https://github.com/quarto-dev/quarto-cli/blob/1fbbc38d211dc23a51f6ed47334063cd56f214ef/src/format/html/format-html-bootstrap.ts#L1923-L1940)

So, should I submit another PR that removes all occurences of this spurious `screen-rightcolumn` specifier? Or is it already taken care of?

dragonstyle:
`screen-rightcolumn` is definitely incorrect FWIW - seems like a typo from long ago! I'll remove that from the schema.

dragonstyle:
https://github.com/quarto-dev/quarto-cli/commit/243d6c1860242d9b3bff05b9c5697420f9653c68 FYI

PR #959, opened 2023-12-29
Update conda env docs

joelostblom:
Follow up from https://github.com/quarto-dev/quarto/issues/341

### DIFF
--- a/docs/projects/virtual-environments.qmd
+++ b/docs/projects/virtual-environments.qmd
@@ -296,7 +296,8 @@ If you need to install Python packages, simply use `pip` or `conda` within the t
 ## VS Code {#vs-code}
 
 If you create a virtual environment in the `env/` directory as described above, Visual Studio Code should automatically discover that environment when you load a workspace from the environment's parent directory.
+Note that this applies only to `venv` not conda environments (which have a separate mechanism for binding to the current VS Code session).
 
-Note that this applies only to `venv` not Conda environments (which have a separate mechanism for binding to the current VS Code session).
+If you have installed the Quarto executable in a conda environment rather than in a system-wide install, you need to launch VSCode from a terminal with that conda environment activated in order for VSCode to detect the conda installation of Quarto. Although there is a setting in the VSCode extension where you can specify the path to the Quarto executable, this for some reason does not work with Quarto installed in conda environments.
 
 You can read more about VS Code support for virtual environments here: <https://code.visualstudio.com/docs/python/environments>.
--- a/docs/projects/virtual-environments.qmd
+++ b/docs/projects/virtual-environments.qmd
@@ -298,6 +298,7 @@ If you need to install Python packages, simply use `pip` or `conda` within the t
 If you create a virtual environment in the `env/` directory as described above, Visual Studio Code should automatically discover that environment when you load a workspace from the environment's parent directory.
 Note that this applies only to `venv` not conda environments (which have a separate mechanism for binding to the current VS Code session via the "Select Python Interpreter" option inside VSCode).
 
-If you have installed the Quarto executable in a conda environment rather than in a system-wide install, you need to launch VSCode from a terminal with that conda environment activated in order for VSCode to detect the conda installation of Quarto. Although there is a setting in the VSCode extension where you can specify the path to the Quarto executable, this for some reason does not work with Quarto installed in conda environments.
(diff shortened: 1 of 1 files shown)


docs/projects/virtual-environments.qmd
```diff
@@ -296,7 +296,8 @@ If you need to install Python packages, simply use `pip` or `conda` within the t
 ## VS Code {#vs-code}
 
 If you create a virtual environment in the `env/` directory as described above, Visual Studio Code should automatically discover that environment when you load a workspace from the environment's parent directory.
+Note that this applies only to `venv` not conda environments (which have a separate mechanism for binding to the current VS Code session).
```
joelostblom, line 299: What separate mechanism is referred to here? Is it "select interpreter"?

docs/projects/virtual-environments.qmd
```diff
@@ -296,7 +296,8 @@ If you need to install Python packages, simply use `pip` or `conda` within the t
 ## VS Code {#vs-code}
 
 If you create a virtual environment in the `env/` directory as described above, Visual Studio Code should automatically discover that environment when you load a workspace from the environment's parent directory.
+Note that this applies only to `venv` not conda environments (which have a separate mechanism for binding to the current VS Code session via the "Select Python Interpreter" option inside VSCode).
 
-Note that this applies only to `venv` not Conda environments (which have a separate mechanism for binding to the current VS Code session).
+If you have installed the Quarto executable in a conda environment rather than in a system-wide install, you need to launch VSCode from a terminal with that conda environment activated in order for VSCode to detect the conda installation of Quarto. Although there is a setting in the VSCode extension where you can specify the path to the Quarto executable, this for some reason does not work with Quarto installed in conda environments.
```
mfisher87, line 301: What do you think about putting this whole thing starting with "Note that ..." and ending with "installed in conda environments" in a callout block? I tried to do a suggestion, but "Applying suggestions on deleted lines is not supported." :( Also, bolding the word **conda** within the callout block may help to draw readers' attention to the piece that's relevant to them.

```markdown
:::{.callout-important}
Stuff about **conda** as an exception to the above instructions here.
:::
```

mfisher87: Maybe even this whole sentence could be bolded :)

>**If you have installed the Quarto executable in a conda environment rather than in a system-wide install, you need to launch VSCode from a terminal with that conda environment activated in order for VSCode to detect the conda installation of Quarto.**

joelostblom: Personally, I don't think a call out would add that much since this section of the page is already short and easy to overview. But I don't feel strongly about not having it either.

cwickham:
I think the VS Code section should mirror the RStudio section and use sub-headings for the different environments: venv, condaenv. We probably also need an `renv` section, but we can handle that.

Then there would seem to be two points in the `condaenv` section:

1. **How to activate the environment.** You mention there is a different mechanism, but the [VS Code docs](https://code.visualstudio.com/docs/python/environments#_create-a-conda-environment-in-the-terminal) seem to imply that it would be automatic? 
    > The Python extension automatically detects existing conda environments. 
   
   If it isn't automatic, then I think there should be a short description on how to do it.

2. **You need to activate the environment a special way if the Quarto CLI is installed inside the condaenv.** The proposed text looks good. But, could we reword the last sentence to: "The Quarto path setting in the Quarto VSCode extension does not work with Quarto installed in conda environments."

joelostblom:
Nice to see you again @cwickham :wave: Thanks for the comments! I updated the last sentence to your suggestion. For the activation of the environment; I believe what VS code automates is the detection of all available environments, but you still need to select which env to use via "Select Interpreter" (as in the screenshot in your link). I'm guessing this is what the original sentence was referring to, but if you think it was just referring to how envs were detected, then we could remove the part I added.

cwickham:
:wave nice to see you too @joelostblom 

Ah, thanks for the clarification regarding detecting vs. activating the environment. I think in that case we should be explicit about that. I'll make a few direct edits.

cwickham:
Preview of updated section: [Guide > Projects > Virtual Environments: Visual Studio Code](https://deploy-preview-959--mystifying-jepsen-fa4396.netlify.app/docs/projects/virtual-environments.html)

cwickham:
Thanks @joelostblom and @mfisher87

PR #960, opened 2024-01-02
Update book-crossrefs.qmd

aborruso:
I have added a premise, which I think is useful, on the use of the '-' symbol.

### DIFF
--- a/docs/books/book-crossrefs.qmd
+++ b/docs/books/book-crossrefs.qmd
@@ -41,6 +41,12 @@ To refer to a section, include a cross-reference to it using an `@` identifier a
 See @sec-introduction for additional discussion.
 ```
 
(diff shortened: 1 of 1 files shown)


cderv:
duplicate of #961 ?

aborruso:
> duplicate of #961 ?

With #961 I have extented #960: I added that the `-` is not needed for a custom prefix.

cderv:
#961 contain commit https://github.com/quarto-dev/quarto-web/pull/960/commits/2d72b49e6136afa29ef5129936b32c0f8214d329 which is also in here. 

So if we merge #961 this PR seems irrelevant

aborruso:
> So if we merge https://github.com/quarto-dev/quarto-web/pull/961 this PR seems irrelevant

Yes, I'm sorry. But I wanted to explain what I had done.

Thank you again

cderv:
All good. Just wanted to be sure I got it right on my side and not close while it should not be. Let's focus on #961 now

PR #961, opened 2024-01-02
Custom prefix per-reference

aborruso:
As we have in [pandoc documentation](https://lierdakil.github.io/pandoc-crossref/#custom-prefix-per-reference), it is not necessary to enter the `-` to have a custom prefix.

### DIFF
--- a/docs/books/book-crossrefs.qmd
+++ b/docs/books/book-crossrefs.qmd
@@ -41,6 +41,12 @@ To refer to a section, include a cross-reference to it using an `@` identifier a
 See @sec-introduction for additional discussion.
 ```
 
+To suppress cross-reference prefix prepend `-` before `@`, like so `[-@citation]`, will suppress default prefix, e.g. `[-@sec-visualization]` will produce just `1.1` (or whatever number this section happens to be) without `Section` prefix.
+
+``` markdown
+See [-@sec-visualization].
+```
+
 To refer to a chapter or appendix explicitly you should spell out "Chapter" or "Appendix" and use the number-only form of cross reference:
 
 ``` markdown
--- a/docs/books/book-crossrefs.qmd
+++ b/docs/books/book-crossrefs.qmd
@@ -50,7 +50,7 @@ See [-@sec-visualization].
 To refer to a chapter or appendix explicitly you should spell out "Chapter" or "Appendix" and use the number-only form of cross reference:
 
 ``` markdown
-See [Chapter -@sec-visualization] for more details on visualizing model diagnostics.
+See [Chapter @sec-visualization] for more details on visualizing model diagnostics.
 ```
 
 


cderv:
Did you test this in a Quarto book ? 

FWIW quarto does not use [pandoc-crossref](https://lierdakil.github.io/pandoc-crossref)  which is not Pandoc documentation but a tool in Pandoc ecosytem. 

Section crossreference in Quarto is its own implementation, and I am surprised we do support `-@sec-*` form. 🤔 

I think we do support what is documented and only the form in brackets. 

The different usage should probably be tested to be completely thorough.

aborruso:
> Did you test this in a Quarto book ?

Yes in [this book page](https://aborruso.quarto.pub/sample/cr.html) (the [code](https://github.com/aborruso/sample/blob/main/cr.qmd)).
 
> FWIW quarto does not use [pandoc-crossref](https://lierdakil.github.io/pandoc-crossref) which is not Pandoc documentation but a tool in Pandoc ecosytem.
> 
> Section crossreference in Quarto is its own implementation, and I am surprised we do support `-@sec-*` form. 🤔
> 
> I think we do support what is documented and only the form in brackets.
> 
> The different usage should probably be tested to be completely thorough.

My goal here is to add two impotant things in quarto doc:

- write how to remove the prefix of cross reference. In the current documentation it is implied;
- in the example where you choose a custom momentary prefix, such as "Chapter" and not "Section," remove `-` because it is not needed.

I hope I have explained myself

cderv:
I see ! 

I guess this related to discussion at
* https://github.com/quarto-dev/quarto-cli/discussions/8068

## Technical context 

Using `[-@sec-intro]` will be parsed as 
```` 
[ Para
    [ Cite
        [ Citation
            { citationId = "sec-intro"
            , citationPrefix = []
            , citationSuffix = []
            , citationMode = SuppressAuthor
            , citationNoteNum = 1
            , citationHash = 0
            }
        ]
        [ Str "(" , Strong [ Str "sec-intro?" ] , Str ")" ]
    ]
]
```

Meaning 
* No prefix used 
* mode is `SuppressAuthor`

In our [current code base (Quarto 1.4)](https://github.com/quarto-dev/quarto-cli/discussions/8068#discussioncomment-8003822), this mean only the number is used. 

Regarding other form with custom prefix `[Custom -@sec-sub]` or `[Custom @sec-sub]`  they are equivalent because 
* `cite.prefix` will be non-empty
* We don't condition on the `SuppressAuthor` in that case. 

It seems all good.

aborruso:
Thank you @cderv

docs/books/book-crossrefs.qmd
```diff
@@ -41,10 +41,16 @@ To refer to a section, include a cross-reference to it using an `@` identifier a
 See @sec-introduction for additional discussion.
 ```
 
+To suppress cross-reference prefix prepend `-` before `@`, like so `[-@citation]`, will suppress default prefix, e.g. `[-@sec-visualization]` will produce just `1.1` (or whatever number this section happens to be) without `Section` prefix.
```
cderv, line 44: Maybe we should mention bracket syntax to be completely exact ... 🤔 

```suggestion
To suppress cross-reference prefix, use bracket syntax and prepend `-` before `@`, like so `[-@citation]`. This will suppress default prefix, e.g. `[-@sec-visualization]` will produce just `1.1` (or whatever number this section happens to be) without default prefix.
```

docs/books/book-crossrefs.qmd
```diff
@@ -41,10 +41,16 @@ To refer to a section, include a cross-reference to it using an `@` identifier a
...
+
+``` markdown
+See [-@sec-visualization].
+```
+
 To refer to a chapter or appendix explicitly you should spell out "Chapter" or "Appendix" and use the number-only form of cross reference:
```
cderv, line 50: If the number-only form is unneeded, mention should be removed. 

```suggestion
To refer to a chapter or appendix explicitly you should spell out "Chapter" or "Appendix" as prefix in the bracket syntax for cross reference:
```

Right ?

cwickham:
Thanks @aborruso! I only added a little context before your edits to allow some re-wording for clarity.

aborruso:
> Thanks @aborruso! I only added a little context before your edits to allow some re-wording for clarity.

That page, the way it was written before, caused me to make several mistakes and waste some time. Also because I am a basic user.

It seems much clearer and more useful to me now. And your contribution is still improving. Thank you

PR #965, opened 2024-01-06
docs: add note about header promotion in book

mcanouil:
Currently, the documentation for books only mention level-one header and YAML title, while in the absence of those, the first header found (independently of its level) will be promoted to title.


cwickham:
@mcanouil This change looks good, but could you remove all the environment stuff from this PR?

mcanouil:
Files wrongly added.
It's fixed.

cwickham:
Perfect, thanks!

PR #971, opened 2024-01-10
Revert "AST processing phases had a change of names (#970)"

cwickham:
I should have merged to `main` since this applies to our currently available highlights.

This reverts commit e253967a9d26b7746ca31c6570cc25a13ec443eb.


cderv:
Oh indeed. I target 1.4 because if was for new feature, but we do have highlight. My bad...

PR #972, opened 2024-01-10
Move Lightbox docs to Guide > Documents > HTML 

cwickham:
Moves the lightbox highlight to a permanent location [Guide > Documents > HTML > Lightbox Figures](https://deploy-preview-972--mystifying-jepsen-fa4396.netlify.app/docs/output-formats/html-lightbox-figures.html):
Screenshot 2024-01-11 at 12 22 45 PM

Questions for @dragonstyle:
* What do you think of the location Guide > Documents > HTML > Lightbox Figures? It feels too long to be a section somewhere, but I'm not really sure it belongs at the same level as HTML Basics, and HTML theming etc. 
* The Reference item is very minimal: 
    Screenshot 2024-01-11 at 12 32 29 PM
    Is that enough, or should we expose the `lightbox` schema?


dragonstyle (approved):
Looks great!

PR #981, opened 2024-01-17
Add example about using include shortcode within source code block

cderv:
and add details for including computations

closes https://github.com/quarto-dev/quarto-cli/issues/8305

### DIFF
--- a/docs/authoring/includes.qmd
+++ b/docs/authoring/includes.qmd
@@ -28,9 +28,9 @@ Include shortcodes need to appear by themselves in a line, and they need to be s
 
 ## Content
 
-A concrete example would be if you have several articles about a topic that share a common introduction. Here we have an article titled "Revealjs Presentations" that wants to include some basic information on presentations not specific to Revealjs (we do that by including `_basics.qmd`):
+A concrete example would be if you have several articles about a topic that share a common introduction. Here, we have an article titled "Revealjs Presentations" that wants to include some basic information on presentations not specific to Revealjs (we do that by including `_basics.qmd`). We also have some demo code stored as scripts that we want to include as non-executed examples (we do that by including `_demo.R` and `_demo.py` inside source code blocks).
 
-``` markdown
+```` markdown
 ---
 title: "Revealjs Presentations"
 ---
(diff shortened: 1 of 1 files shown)


docs/authoring/includes.qmd
```diff
@@ -28,9 +28,9 @@ Include shortcodes need to appear by themselves in a line, and they need to be s
 
 ## Content
 
-A concrete example would be if you have several articles about a topic that share a common introduction. Here we have an article titled "Revealjs Presentations" that wants to include some basic information on presentations not specific to Revealjs (we do that by including `_basics.qmd`):
+A concrete example would be if you have several articles about a topic that share a common introduction. Here, we have an article titled "Revealjs Presentations" that wants to include some basic information on presentations not specific to Revealjs (we do that by including `_basics.qmd`). We also have some demo code stored as scripts that we want to include as non-executed examples (we do that by including `_demo.R` and `_demo.py` inside source code blocks).
```
cwickham, line 31:
```suggestion
A concrete example would be if you have several articles about a topic that share a common introduction. Here, we have an article titled "Revealjs Presentations" that wants to include some basic information on presentations not specific to Revealjs (we do that by including `_basics.qmd`). We also have some demo code stored as scripts that we want to include as non-executed examples (we do that by including `_demo.R` and `_demo.py` inside source code blocks):
```

docs/authoring/includes.qmd
```diff
@@ -64,16 +82,32 @@ title: "My Document"
...
+
+````
+
 A couple of important things to remember when using computational includes:
 
 1)  All computations still share a single engine (e.g. knitr or jupyter)
 
 2)  Computational includes work only in `.qmd` files (they don't work in `.ipynb` notebook files)
 
+Note that you can't use `include` shortcode within a computational code block itself - as the example above shows, the executable code block needs to be inside the included document.
```
cwickham, line 105:
```suggestion
Note that you can't use  the `include` shortcode within a computational code block itself - as the example above shows, the executable code block needs to be inside the included document.
```

cwickham (approved):
This looks great thanks @cderv

PR #993, opened 2024-01-26
Add initial version of blog post (Connect email)

rich-iannone:
This adds a post that walks the user through two simple use cases for Quarto emailing via Connect.

### DIFF
--- a/docs/blog/posts/2024-02-30-connect-email/index.qmd
+++ b/docs/blog/posts/2024-02-30-connect-email/index.qmd
@@ -1,0 +1,170 @@
+---
+title: Quarto Emailing Support in Posit Connect 
+subtitle: This new feature gives you the power to craft informative emails and specify the delivery of them through Connect.
+description: |
+  Quarto 1.4 now has support for sending email through Posit Connect as part of a larger HTML report. We'll walk you through how this all works with numerous examples.
+categories:
+  - Email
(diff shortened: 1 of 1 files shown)


cderv:
Hey @rich-iannone, should we published it ? 

I'll talk about email next week in a Posit Workflow Demo with a youtube video. Could be a good idea to sync. I'll review this to see if there is more to include - or maybe reference to the Youtube video I'll do.

cwickham:
I'd talked to Rich in Slack about:

> I think you should lead with a bit more of the end result before diving into the details of implementation. I.e. many readers might not have Connect (so they probably don't care about the details), but this is a good opportunity to show them a really cool reason to have Connect. That could look like some screenshots (maybe the rendered email, and something that shows the scheduling capacity?), or maybe a side by side of the source doc and a rendered email.".

I think a link to the YouTube video would basically cover my comments. I.e., make it more of an announcement: you can use Quarto + Connect to send emails; here's a video demoing it; and here's where to read more.

cderv:
@rich-iannone @cwickham We should probably do something with this blog post 

I am starting to talk about Email in this video: https://youtu.be/Hl9TvhlzfV4?t=1400

Not the best quality and not just on email. If we want better video content, Rich you could probably do a quick one with Greg to demo the current feature if we want to add to the blog post

PR #997, opened 2024-01-31
Update email from rstudio to posit

randyzwitch:
Confirmed with Kim B. that this alias does exist for Posit, so changing to keep branding consistent


randyzwitch:
Also exists here, possibly others

https://github.com/quarto-dev/quarto-web/blob/1d928ed9440955f61f25399d0123e2ef068f2065/about.qmd#L68

cwickham:
Thanks @randyzwitch !

PR #999, opened 2024-02-02
Add raw blocks and spans

cwickham:
Closes https://github.com/quarto-dev/quarto-cli/issues/4500

### DIFF
--- a/docs/authoring/markdown-basics.qmd
+++ b/docs/authoring/markdown-basics.qmd
@@ -208,6 +208,29 @@ code
 
 If you are creating HTML output there is a wide variety of options available for code block output. See the article on [HTML Code](../output-formats/html-code.qmd) for additional details.
 
+## Raw Blocks and Spans
+
+Use raw blocks to include content directly without Quarto parsing it. A raw block starts with ` ```{=` followed by a format and closing `}`, e.g. here's a raw HTML block:
+
+```` markdown
+```{=html}
+<iframe src="https://quarto.org/" width="500" height="400"></iframe>
+```
+````
+
+For PDF output use a raw LaTeX block:
+
+```` markdown
+```{=latex}
+\renewcommand*{\labelitemi}{\textgreater}
+```
+````
+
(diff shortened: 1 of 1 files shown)


cwickham:
@cderv I've given raw HTML and LaTeX examples because I think they are the most common. Are there others that might be worth calling out?

docs/authoring/markdown-basics.qmd
```diff
@@ -208,6 +208,29 @@ code
 
 If you are creating HTML output there is a wide variety of options available for code block output. See the article on [HTML Code](../output-formats/html-code.qmd) for additional details.
 
+## Raw Blocks and Spans
```
cderv, line 211: Should it be more generic ? 

```suggestion
## Raw Content
```

To be precise, the syntax for inline is not really **Spans** - I don't know what Spans would really means. It is just the other Spans that we usually talk about is Divs & Spans and the syntax is different 
```
[This is *some text*]{.class key="val"}
```
From `brackted_spans` extensions (https://pandoc.org/MANUAL.html#extension-bracketed_spans)

Here Pandoc's manual talks about 
> Inline spans and fenced code blocks with a special kind of attribute

that is not much clearer... 

You used below 
* raw span
* raw block

So I understand the title of Spans, but just thinking about confusion for users with Spans from https://quarto.org/docs/authoring/markdown-basics.html#divs-and-spans

You know better for good understanding, just sharing my thoughts.

cderv (approved):
> Are there others that might be worth calling out?

Maybe typst example for less classical format ? 

Question I have: Should we link to Pandoc documentation about those features ? 
* https://pandoc.org/MANUAL.html#extension-raw_attribute

They are Pandoc feature, available Quarto because directly working with Pandoc. This would. 

Otherwise, all goo

docs/authoring/markdown-basics.qmd
cwickham: Yes, I'd definitely confused them. Fixed it to not talk about divs or spans.

PR #1002, opened 2024-02-04
Add callout to not use inline code in YAML

cwickham:
Closes https://github.com/quarto-dev/quarto-cli/issues/6389


cscheid (approved):
Looks good. Would it make sense to repeat the second callout in the crossreferences authoring section, since it's likely that people might get the idea while reading that one too?

PR #1010, opened 2024-02-08
Update Website Listings

cwickham:
Closes https://github.com/quarto-dev/quarto-cli/issues/8303

- [x] ~Add support for programmatically filtering content from a listing using `include` or `exclude` with glob syntax to include or exclude specific items from the listing. See <https://github.com/quarto-dev/quarto-cli/commit/d415d9ca5b7cb59a8a4750dd3eeb60116b931bd6s>~ Already exists at https://quarto.org/docs/websites/website-listings.html#including-or-excluding-items
- [x] ([#6407](https://github.com/quarto-dev/quarto-cli/issues/6407)): Add supporting for the field `word-count` for listing items. It is not displayed by default.  **Added entry in [fields table](https://deploy-preview-1010--mystifying-jepsen-fa4396.netlify.app/docs/websites/website-listings.html#listing-fields) in Guide**
- [x] ~([#4800](https://github.com/quarto-dev/quarto-cli/issues/4800)): Add support for including an `xml-stylesheet` in listings. Use the `xml-stylesheet: example.xsl` under `feed:` to provide a path to an XSL style sheet to style your RSS feed.~ **In reference.**
- [X] ([#7290](https://github.com/quarto-dev/quarto-cli/issues/7290)): Add support for `feed:type` of `metadata`, which will use only explicitly provided description metadata when generating an RSS feed. Additionally, note that `partial` feed types will prefer to use an explicit description over the first paragraph, when a description is available. **Edited entry in [feeds table](https://deploy-preview-1010--mystifying-jepsen-fa4396.netlify.app/docs/websites/website-listings.html#feeds) in Guide**


dragonstyle (approved):
Also :fire:

PR #1030, opened 2024-02-27
Fix internal links using `.qmd`

cderv:
They need to be internal link to support `.html` transformation correctly.

This fixes https://github.com/quarto-dev/quarto-cli/issues/8902

As noted by @mcanouil, there is quite a number of links that are using absolute links starting with `https://quarto.org` instead of internal links started with `/docs/...`

@cwickham we are not sure if we are missing something regarding usage of absolute links. Probably internal links could be used instead (so that they work correctly locally or on preview for example). 

@mcanouil offered to make a PR is needed. 

Thanks

### DIFF
--- a/docs/websites/_page-navigation.md
+++ b/docs/websites/_page-navigation.md
@@ -17,7 +17,7 @@ page-navigation: false
 ---
 ```
 
-Or to control page navigation for all pages in a directory specify `page-navigation` in [`_metadata.yml`](https://quarto.org/docs/projects/quarto-projects.qmd#directory-metadata):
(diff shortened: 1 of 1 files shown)


cderv:
Link is now fixed at https://deploy-preview-1030--mystifying-jepsen-fa4396.netlify.app/docs/books/book-structure.html#page-navigation

mcanouil:
FYI, using only internal reference to `qmd` file will change 63 files.

cwickham:
This looks good. I'll merge now so this link is fixed, and @mcanouil you can do a new PR if you'd like to catch all the others.

cwickham (approved):

mcanouil:
It's ongoing.

- #1031

PR #1051, opened 2024-03-15
Fix leaflet examples in `article-layout.qmd`

cwickham:
[Preview](https://deploy-preview-1051--mystifying-jepsen-fa4396.netlify.app/docs/authoring/article-layout.html)

PR #1052, opened 2024-03-15
Re render docs\authoring\article-layout.html 

cderv:
with current renv status so that freeze as correct information

Would close #1051 as an alternative. 

This is the missing frozen file from #957 rendered with current version of R packages in the `renv.lock`


cderv:
This works ok https://deploy-preview-1052--mystifying-jepsen-fa4396.netlify.app/docs/authoring/article-layout.html#screen-column

PR #1056, opened 2024-03-20
Update RMarkdown <-> Quarto table to include Quarto Dashboards

kmasiello:
Updated the FAQ for RMarkdown to replace the flexdashboard equivalent from "planned" to "Quarto Dashboards"


cderv:
Oh good catch ! thanks !

PR #1057, opened 2024-03-22
Add callout about HF running badge

gshotwell:
I ran into some issues where I thought the HF build process was complete when it was still in a "Building" state. There's not much we can do about the core issue, but this at least adds some additional warnings.


cderv:
Thank you !

PR #1064, opened 2024-04-02
fix: image display issue on mobile

mcanouil:
This pull request fixes the issue where an image with a fixed height leads to bad display. The changes include updating the `index.qmd` file to remove the fixed height attribute for the image and setting a `max-height` instead.


mcanouil:
I double checked using the preview. The image no longer overflow and no obvious changes on wide display.
So all seems good.

PR #1070, opened 2024-04-07
[DOC] small fixes

yogabonito:


cderv (approved):
Thanks you ! We need to re-render as this file hs frozen state. I'll do it.

PR #1072, opened 2024-04-09
Add note on changing the alt text behavior with backslash

joelostblom:
Follow up from https://github.com/quarto-dev/quarto-cli/discussions/9310. I think it is helpful to document this behavior as it is the default in many other markdown specs and could help transitioning existing material to quarto.


cscheid (commented):
Please fix:

- replace "pandoc" with "Pandoc"
- replace "the text within square brackets as the alt text instead of as the caption" with "the caption instead as alt text"

joelostblom:
@cscheid Thanks for the quick comments, updated!

cscheid (approved):

PR #1090, opened 2024-04-26
fix: aliases use qmd instead of html

mcanouil:
This pull request fixes the file extensions of the aliases in the documentation files. The aliases were previously using the .qmd extension instead of the correct .html extension.

Fixes quarto-dev/quarto-cli#9494


aronatkins (approved):

cderv:
Thanks !

PR #1096, opened 2024-05-07
Add search to prerelease

cwickham:


_quarto.yml
```diff
@@ -23,7 +23,7 @@ website:
     show-item-context: true
     type: overlay
     algolia:
-      index-name: prod_QUARTO
+      index-name: prerelease_QUARTO
```
cderv, line 26: I think this one we need to keep in the prerelease branch, and switch back to the right index when we'll merge. 

Unless you have other idea. 🤔 

I don't know if we support env var setting for this (instead of YAML config). If we support YAML we could just set the env during the rendering.

.github/workflows/upload-index.yml
cderv, line 1: I would use two steps to update both site at once. Like the update-downloads workflow. 

Also because I don't think the cron schedule will really trigger from a branch. I do think this is only considered from main  branch. 

Not sure which part triggers the repository dispatch though... 🤔

.github/workflows/upload-algolia.rb
cderv, line 1: I would try to use parameter here, or an env var with a default. 

Otherwise, we'll need to handle merge conflict when merging prerelease into main (if this is our planned workflow for release doc update)

cderv:
@cwickham I have updated the PR. Does it make sense ? 

Is there a scenario where we want to update only one index and not the other ? I did not make any condition on that. Not sure it was worth it, but it could be done.

.github/workflows/upload-index.yml
```diff
@@ -27,5 +27,16 @@ jobs:
           ALGOLIA_APP_ID: ZPJB5I1QN7
           ALGOLIA_INDEX: prod_QUARTO
           QUARTO_INDEX_PATH: search.json
+          QUARTO_INDEX_URL: https://prerelease.quarto.org/search.json
```
cwickham, line 30: This should be the released site, right?
```suggestion
          QUARTO_INDEX_URL: https://quarto.org/search.json
```

_quarto.yml
cwickham: I think you are right. Is this something we could use profiles for?

.github/workflows/upload-index.yml
cderv: 😓 yes ! 

Thanks for the review !

_quarto.yml
cderv: Yes definitely. If you plan to use profile for other settings in `_quarto.yaml` we should move it. 

I tihnk this is a great idea. We can use profile for prerelease for this website to adapt some config, and then render the prerelease version in the branch using `--profile` argument or setting the environment variable.

cderv: Maybe we can merge this and then deal with that later depending on the best workflow you'll came up with

PR #1099, opened 2024-05-17
document confluence env vars

hutch3232:
I could not find which Confluence environmental variables are valid. I wasn't sure where it was most appropriate to document them. I put them on the CI page, but perhaps the Confluence page would be even better. I chose CI since there were other env vars there. 

This was the only location where I found these env vars defined: https://github.com/quarto-dev/quarto-cli/discussions/6915

### DIFF
--- a/docs/publishing/ci.qmd
+++ b/docs/publishing/ci.qmd
@@ -184,3 +184,11 @@ export CONNECT_SERVER=https://connect.example.com/
 export CONNECT_API_KEY=7C0947A852D8
 quarto publish connect --id DDA36416-F950-4647-815C-01A24233E294
 ```
+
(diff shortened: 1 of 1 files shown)


cderv (commented):
Thanks ! This should definitely be documented. It is tracked at 
* https://github.com/quarto-dev/quarto-cli/issues/6923

We currently show in action examples for now
https://github.com/quarto-dev/quarto-actions/blob/a3b9e3b6268d34e343e9bdb13ba3dfe789f278ec/examples/quarto-publish-example.yml#L71-L79

So we could add also an example I guess if we add to the table. Can you add this ?

hutch3232:
@cderv Added an example, hope that is what you were looking for. Thanks!

PR #1101, opened 2024-05-20
Pre-release Banner

cwickham:
Attempt at adding a banner for pre-release docs site.

Uses a profile, so to see result use:
```
quarto preview --profile prerelease-docs
```

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -49,22 +49,26 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-            if [ "${{ github.event.pull_request.head.ref }}" == "prerelease" ]; then
+            if [ "${{ github.event.pull_request.base.ref }}" == "prerelease" ]; then
               echo "is_prerelease_docs=true" >> "$GITHUB_OUTPUT"
-            elif [ -n "${{ fromJSON(github.event.issue.pull_request) }}" ]; then
+            elif [ -n "${{ github.event.issue.pull_request.url }}" ]; then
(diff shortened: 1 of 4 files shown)


cwickham:
@cderv This PR adds a banner to the pre-release site via a profile. We should talk about what that means for CI. 

One option would be to always render (for previews, and deployments) the pre-release docs site with the `pre-release-docs`.

I notice we also use profiles to control the behavior of the download page. We should probably insulate the pre-release docs site from that, so it always uses the same profile.

cderv (approved):
It seems we have issue on Actions right now, but this PR is now ready on my side. I think the GHA part will work.

cderv (approved):
All good ! Let's merge !

cderv:
@cwickham Idea: Should we add a link to https://prerelease.quarto.org in the PreRelease download page from https://quarto.org ? 

At least here: https://quarto.org/docs/download/prerelease.html
Or in the https://quarto.org/docs/download/ page in the pre-release tab ? 

Or otherwise, more prominent in the Nav bar somewhere. 

I am wondering on how to make this findable

cwickham:
#1106  Adds a link to the `_highlights.yml` which [does show up on the download page](https://deploy-preview-1106.quarto.org/docs/download/prerelease) but near the bottom. It may be worth making it more obvious, what do you think?

cderv:
Oh great! For this place, it may be enough. 

I think I am more generally thinking we may want a menu item to easily switch between both when you consult one (instead of modifying the url in browser)

cwickham:
Oh yeah, that's a great idea. Do you have an idea on how to implement that?

PR #1127, opened 2024-06-03
Pre-release `_quarto` config

cwickham:
Some additional edits to distinguish from main site


_quarto-prerelease-docs.yml
```diff
@@ -2,6 +2,11 @@
 version: v1.5
 
 website:
+  title: "Quarto (Pre-release)"
+  google-analytics: "G-FV9Z7SDZ0M"
```
cwickham, line 6: @cderv This is the same as in the main site, but it probably shouldn't be? Or do we need to add something to distinguish this site on GA?

_quarto-prerelease-docs.yml
```diff
@@ -2,6 +2,11 @@
...
+  image: "quarto-dark-bg-pre.jpeg"
+  site-url: https://prerelease.quarto.org
```
cwickham, line 8: @cderv We need to change `site-url` right? It feels like it might bite us if it is the same as `quarto.org`?

_quarto-prerelease-docs.yml
```diff
@@ -2,6 +2,11 @@
...
+  repo-url: https://github.com/quarto-dev/quarto-web/tree/prerelease
```
cwickham, line 9: @cdrev This isn't right. The "Edit this page" links won't work. But, I don't know what to put here so they do work.

_quarto-prerelease-docs.yml
cderv: Oh good point. Do you have access to this ? I don't know how we messed up the data 😅 

@icarusz do you have access ? probably worth adding @cwickham ? 

Or should we switch to other tool ?

_quarto-prerelease-docs.yml
cderv: yes definitely. I think this has impact on some feature like socials and RSS. So it seems ok to change !

_quarto-prerelease-docs.yml
cderv: We need `repo-branch` here: 
https://quarto.org/docs/websites/website-navigation.html#github-links

So main `repo-url`, but adding `repo-branch` so that main is not used by default. 

Reminder: the auto cherry pick happen from main to prerelease, so if a typo is spotted on prerelease, as PR reviewer, we'll need to make sure this is not to merge in `main` so that fixed get in both place. 

I think that is ok, better than a "edit this page" link that may not exists (if our prerelease website has pages that our main website does not have)

cderv (approved):
Thanks for thinking about all this !

_quarto-prerelease-docs.yml
cwickham: I discussed with @icarusz  in Slack...going to leave the same as quarto.org for now.

PR #1134, opened 2024-06-06
Gallery

cwickham:


docs/gallery/gallery.ejs
cderv, line 1: Following your question slack, what I tried was 

- Setting `overflow-hidden` on the `div.card`
- Not setting rounded bottom on the img. 

I believe this could remove the white gap. I did not push to the PR to update the preview, but in browser developer mode this works.

PR #1153, opened 2024-06-11
docs: add task list example

mcanouil:


cderv (approved):
I think we can add it to main website... so PR on main. 

This is note a new 1.5 feature support, right ?

mcanouil:
It's really a missing example in list. Nothing more.

As I don't know what is the main/pre-release workflow exactly, I assumed new content goes to pre-release and main is for "typos".

Is prelease only for new things/featurzs in CLI? Everything else goes into main?

cwickham:
Yup, `prerelease` is primarily for new features. General doc updates and improvements can go to `main`. For PRs merged to `main`, we have a workflow that automatically creates a PR against `prerelease` to keep it updated.

mcanouil:
Done. New target is `main`.

PR #1159, opened 2024-06-13
Update website-listings-custom.qmd

andrewpbray:
Fix what appears to be a typo (the example that follows uses `templateParams`instead of `params`).


cderv:
😆 I pointed you to that doc in an issue, I have read it, and missed the typo ! 

Thanks !

PR #1165, opened 2024-06-18
Update citation.json

s2t2:
I guess you can use the `id` parameter to customize the identifier that shows up at the top. 

Screenshot 2024-06-18 at 1 11 16 PM

Adding this info to the docs.

### DIFF
--- a/docs/reference/metadata/citation.json
+++ b/docs/reference/metadata/citation.json
@@ -4,6 +4,10 @@
     "title": "Citation",
     "options": [
       {
+        "name": "id",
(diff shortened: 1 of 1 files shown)


cwickham:
Hi @s2t2. Thanks for pointing this out. These `.json` files are actually automatically generated from the [schema](https://github.com/quarto-dev/quarto-cli/tree/main/src/resources/schema). So, rather than make a change to this file, we either need to fix something in `tools/reference.ts` or something in the schema. 

I think this is tracked in: https://github.com/quarto-dev/quarto-cli/issues/8713 

I'll close this PR in favor of that issue.

PR #1167, opened 2024-06-18
Add example of customizing format-links

cwickham:
Adds example in [Guide > HTML > Including Other Formats](https://deploy-preview-1167.quarto.org/docs/output-formats/html-multi-format.html#specifying-formats-to-link)

Closes https://github.com/quarto-dev/quarto-cli/issues/8610

### DIFF
--- a/docs/output-formats/html-multi-format.qmd
+++ b/docs/output-formats/html-multi-format.qmd
@@ -70,6 +70,29 @@ format:
 format-links: [ipynb]
 ```
 
+You can also customize the links by passing `format-links` an object with keys:
+
+* `format`, the format to be linked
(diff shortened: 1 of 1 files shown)


cderv (commented):
LGTM. 

Do we have also somewhere doc for the other schema allowed ? 
````
- text: Other Link
    href: https://quarto.org/
    icon: hand-thumbs-up
````

Otherwise we could add example here like I shared in the issue. 

Basically, it allows to add external link the this sections.

cwickham:
Oh, good point. We do document that schema here: https://quarto.org/docs/output-formats/html-basics.html#code-links-and-other-links

I'll add an example and link to that section.

PR #1171, opened 2024-06-19
Note that Quarto doesn't follow project-wide engine yet

alecloudenback:


cderv:
Thanks ! 

Linking to https://github.com/quarto-dev/quarto-cli/issues/3157 so that we remember to update

PR #1174, opened 2024-06-20
docs: add a note about `width` (in column) being not supported in `pptx`

mcanouil:
Fixes quarto-dev/quarto-cli#2635


cwickham:
It seems weird to document something on a "`pptx` only" page, then immediateily say it doesn't work in `pptx`. I think a cleaner option here is to use conditional content and show an example without `width`. I added a commit to do this.

PR #1181, opened 2024-06-24
Update website-navigation.qmd to use `_index.qmd`

jameslairdsmith:
Two tweaks here would have helped me understand the "auto" option for creating navigation in websites a little bit better.

- Using `_index.qmd` instead of `index.qmd` because the latter creates its own page, which is often not what you want.
- Telling the user they can also control the order of a sub-directory using `_index.qmd`.

### DIFF
--- a/docs/websites/website-navigation.qmd
+++ b/docs/websites/website-navigation.qmd
@@ -186,9 +186,9 @@ Using `contents: auto` at the root level will result in all documents in your we
 
 1.  Navigation item titles will be read from the `title` field of documents.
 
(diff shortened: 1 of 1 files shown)


cderv:
Thanks. I am suprised by this `_index.qmd` usage. It is not something I was aware of it explicitly working. I would like to check this usage, because as soon as we document it it will be hard to change, and for now I wonder if this is something link to other thing. 

Using a `_` prefix in name like `_index.qmd` should mean that this document is excluded from render list for the project, so I am surprised setting metadata in it will be catched.

PR #1192, opened 2024-06-25
Update custom-formats.yml

mvuorre:
Add link to https://github.com/mvuorre/quarto-preprint

Only question is where in the list I should add this? Felt a bit cheeky putting it on top but OTOH it is the latest addition.


cderv:
Thanks

PR #1209, opened 2024-06-27
Add Typst details to main citation page

cwickham:
[Preview](https://deploy-preview-1209.quarto.org/docs/authoring/citations.html)
Closes https://github.com/quarto-dev/quarto-cli/issues/9762

### DIFF
--- a/docs/output-formats/_typst-citations.md
+++ b/docs/output-formats/_typst-citations.md
@@ -1,0 +1,25 @@
+Typst comes with its [own citation processing system for bibliographies](https://typst.app/docs/reference/model/bibliography/) and using `format: typst` defaults to it. To specify a bibliography style using Typst's system, use the `bibliographystyle` option. Provide a string from [Typst's list of built-in styles](https://typst.app/docs/reference/model/bibliography/#parameters-style), e.g.:
+
+  ``` yaml
+  bibliography: refs.bib
+  bibliographystyle: apa
+  ```
+  
+  Or alternatively, provide a path to a local CSL file:
+  
+  ``` yaml
+  bibliography: refs.bib
+  bibliographystyle: my-csl-style.csl
+  ```
+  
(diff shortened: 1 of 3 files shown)


cwickham:
@cderv This adds the Typst details to the main citation page and calls out that Typst is different at the top.  I think it still may be confusing since it isn't clear what parts of the page still apply to Typst (I think none, except "Citation Syntax" and I guess "Bibliography Files" although the supported formats might be different). Is it worth a bigger re-organize to clear that up, or is what is in this PR good enough.

docs/output-formats/_typst-citations.md
cderv, line 1: This part could be simplified later when we'll "fix" our missing support for CSL that pandoc's support 

- https://github.com/quarto-dev/quarto-cli/issues/10168

This means also that Citation Style will apply.

cderv (approved):
> This adds the Typst details to the main citation page and calls out that Typst is different at the top. I think it still may be confusing since it isn't clear what parts of the page still apply to Typst (I think none, except "Citation Syntax" and I guess "Bibliography Files" although the supported formats might be different). Is it worth a bigger re-organize to clear that up, or is what is in this PR good enough.

I think it is good enough, but I understand the unclear state. Typst is I think the only format we have supporting citeproc but not defaulting to it. (Maybe we should change that 🤔 ) 

Probably a reorganization would be helpful. A simple one could be 

- This pages is about Citations. We could bigger sub part to clearly explain the different system that could happen
- One about Pandoc's citeproc built in system. Almost all content of the current page would be under it
- One about Format specific system and 
	- A subpart about LaTeX system
	- Another about typst own system

The pandoc syntax could be at the top as a common part. 

Not urgent though, it works well like this.

cwickham:
@cderv Thanks for the feedback. I opened https://github.com/quarto-dev/quarto-cli/issues/10204 to track these suggestions and will merge this.

PR #1213, opened 2024-06-27
Set Algolia index in prerelease-docs profile

cwickham:
I think I put the `algolia: index-name` in before we were using profiles. This undoes the change in `_quarto.yml` and instead sets it in `_quarto-prelease-docs.yml`:

```
website:
  search:
    algolia:
      index-name: prerelease_QUARTO
```

### DIFF
--- a/_quarto-prerelease-docs.yml
+++ b/_quarto-prerelease-docs.yml
@@ -11,6 +11,9 @@ website:
     content: "Pre-release ({{< meta version >}}) Documentation: [Download {{< meta version >}}](/docs/download/prerelease.qmd), [Current Release Docs](http://quarto.org) "
   navbar:
     pinned: true
+  search:
+    algolia:
+      index-name: prerelease_QUARTO
 
 format:
   html:


cwickham:
@cderv Just adding you as a reviewer to double-check this has no consequences for the way actions are set up.

cderv:
Oh so it means our search was broken ? 🥹 

I don't know much about algolia, but checking our workflow, I do think I am not doing the right thing... 

The search.json that needs to be updated to the prerelease index on algolia should be the one from the prerelease branch ... right ? 

Currently, I think we push the one from main in both case 🤦 

https://github.com/quarto-dev/quarto-web/blob/ac6997088542998def4eb51b1ceac49a54a210be/.github/workflows/upload-index.yml#L21-L40

If you confirm we can tweak this PR to adapt the workflow too.

cderv:
I changed the base branch to be `main` since we have merged prerelease and this should be done in main with the other changes

cwickham:
As far as I know the search was working. Doesn't the combination of:
```
ALGOLIA_INDEX: prerelease_QUARTO 
```
and 
```
QUARTO_INDEX_URL: https://prerelease.quarto.org/search.json 
```
ensure the right file is uploaded to the right index?

The change in this PR is just to ensure the search on `prerelease.quarto.org` hits the right index.

cwickham:
To clarify, `prerelease.quarto.org` was hitting the right index, but because it was using the value from `index-name` from `_quarto.yml`, but it seems like it makes more sense to make that change on the `prerelease-docs` profile.

cderv:
You're right this is all good

PR #1215, opened 2024-06-28
Typst CSS, Raw HTML, and Typst snapshots, take 2

gordonwoodhull:
This replaces #1162[^1], and fixes https://github.com/quarto-dev/quarto-cli/issues/9881

[^1]: which I accidentally closed by force-pushing `prerelease` to the PR branch, so GitHub thinks it is merged. 😆 

In addition to the documentation changes, this adds `pre-render` invocations in `_quarto.yml` for two snapshots of Typst tables, closing quarto-dev/quarto-cli#10118.

We can add another script here to organize more snapshots.

I have tested it in CI up to the point of it [needing `magick`](https://github.com/quarto-dev/quarto-web/actions/runs/9717256246/job/26822571267?pr=1162). @cderv please add Image Magick to the configuration when you have a chance.

Although the freezes that generate the snapshots are checked in, I'm not currently checking in the snapshots. I think what needs to happen is that they are checked in but we periodically delete and regenerate them in order to receive updates in response to package changes, changes in quarto, etc.


gordonwoodhull:
still nothing lol, github hates me.

gordonwoodhull:
This no longer has the problematic pre-render hooks, and IMO everything is ready. Merging.

PR #1219, opened 2024-06-30
Blog - beautiful tables in Typst

gordonwoodhull:
I have most of the mechanics for the Typst CSS blog post, a bunch of examples, and a few words.

However, I did some less-than-perfectly-reproducible things in order to get this working:
- Needed _different_ packages installed in order for Great Tables to run, created local `.venv` for it. Something about numpy versions.
- I need to render minimal HTML examples without any website chrome in order to embed them in iframes, because of the mix of languages. Not knowing how to disable the project and do a local build, I instead created a script that touches `_quarto.yml` before render, and removes it after. Surely there is a better way!
- The examples are cluttered with various codes to turn off non-essential output for the snapshot and for the iframe. Some of this could be done from the command line but not all (and I couldn't find a way to turn off the title except `-M title:false`).
- manual iframe sizing
- centering each within a 50% column seems to look okay, still needs nudge with `padding-top` on the Typst snapshot, since that one is generally wider/shorter

Ignoring these warts, it's pretty easy to add new examples and at least awkwardly reproducible.


cwickham:
> I need to render minimal HTML examples without any website chrome in order to embed them in iframes, because of the mix of languages. Not knowing how to disable the project and do a local build, I instead created a script that touches _quarto.yml before render, and removes it after. Surely there is a better way!

I think making sure they aren't in the render list works. E.g. putting them inside a folder starting with `_` should mean you can render them without all the website stuff, although now you need to render explicitly, or in a separate render step ... 🤷🏻‍♀️ not sure if that helps or not.

gordonwoodhull:
Thanks @cwickham, it's good to know that `_` works with a folder. 

I puttered around with removing the individual examples from the render list but in the end I wanted full examples with code to link to, so I end up rendering every example 4 times.

I made the blog post into a listicle, check it out here: 

https://deploy-preview-1219.quarto.org/docs/blog/posts/2024-07-02-beautiful-tables-in-typst/

gordonwoodhull:
Okay, I'm merging this before I lose another day.

Glad to revise if there are issues.

PR #1222, opened 2024-07-01
gha - add imagemagick to the build

cderv:
@gordonwoodhull Let's just add image magick to preview and see if this help you branch building. 

This is how I would install it. If that works ok, we need to add the same step of other relevant workflow. 

FWIW This PR is not against main, but against your branch needing Magick.

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -62,6 +62,12 @@ jobs:
         with:
           version: ${{ steps.github-release.outputs.version }}
 
+      # Add a step to install image magick on Linux only
+      - name: Install ImageMagick
+        run: |
+          sudo apt-get install imagemagick
+          which magick
+
       - name: Is it for prerelease website ?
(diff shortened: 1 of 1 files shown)


cderv:
It built, but using the workflow from main (🤦 obviously). So we'll test this once you merge in your branch maybe

gordonwoodhull:
After our discussion, I'm not 100% convinced that we want `magick` on main. 

The way I set this up, if the snapshots are there, the script is a no-op. Maybe we want to keep the snapshots checked in, and have a workflow that has `magick` _plus_ all the table libraries, that force-regenerates the snapshots. Then run that weekly or whatever.

gordonwoodhull:
I've copied this branch locally for reference, thank you!

But as we discussed, I think it's the wrong idea and we should have a separate workflow for this. I'll close this and open another issue for discussion.

cderv:
Ok great. Let's do this !

PR #1250, opened 2024-07-08
Add pages for release notes

cwickham:
A rather manual take on having release notes in permanent locations on quarto.org of the form `quarto.org/docs/download/changelog/1.*/` (e.g. [1.4](https://deploy-preview-1250.quarto.org/docs/download/changelog/1.4/)). 

Available via links from "Older Releases" tab. Current and prerelease are aliased to their respective pages.

We would have to add the following tasks to the release checklist:

* Copy `changelog` from old release to `_changelog.md` in `docs/download/changelog/1.x`

* Copy `docs/download/changelog/1.*/index.qmd` and increment versions

* Add new item to `download-older` listing in `docs/download/index.qmd` 
  
* Increment version in `aliases` of:
   - `docs/download/release.qmd`
   - `docs/download/prerelease.qmd`

Would close https://github.com/quarto-dev/quarto-cli/issues/5396

### DIFF
--- a/docs/download/changelog/1.1/_changelog.md
+++ b/docs/download/changelog/1.1/_changelog.md
@@ -1,0 +1,158 @@
+## Jupyter
+
+- Daemonize jupyter notebooks referenced within listings (e.g. in a blog)
+- Refine over-detection of Jupyter shell magics (which prevented kernel daemonization)
+- Use on-disk cache for filtered jupyter notebooks
+- Prevent error in `quarto check` when Jupyter is installed but has no Python kernel
+- Don't run `ipynb-filters` for qmd source files (only run them for ipynb source files)
+- More gracefully handle cell outputs with no data (don't print warning, just ignore)
+- Handle non-textual data from jupyter's plain text output more robustly (#1874)
+- Use IJulia's built-in conda environment / jupyter install for julia notebooks/qmds
+
+## Knitr
+
+- Correct handling of `knitr::include_graphics()` within inline expressions.
+- Improve error message for HTML being emitted in non-HTMl formats (#1745)
+- Compatibility with rgl plots (#1800, thanks @dmurdoch)
+
+## OJS
+
+- Better handle OJS code blocks that begin with empty lines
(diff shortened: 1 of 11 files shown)


cwickham:
@cderv, this is a pretty manual approach but is designed to solve the issue of not having places on quarto.org for release notes. I imagine there are more automated approaches, but maybe a once a release edit isn't too bad?

cderv:
I think as long as we have the manual checklist for release this is ok. 

Though we could also do simple automation: 

- Have a script (that we could also run with `quarto run`, to do exactly the step you mentioned  (that also play the role of documentation)
- I don't think having some CI automation for this is worth the effort and maintainance for once per version. At least until we'll do other manual step. 

What do you think about the script ? 🤷‍♂️

cwickham:
@cderv Yeah, a script could be good. So, the manual step becomes: run `update-release-notes.sh`, or something. Is this something we should do in this PR? Or should we merge this, then figure out the script?

I have a slight preference for earlier merging so I can link to the 1.5 release notes in the blog post I'm working on for this week.

cderv:
> So, the manual step becomes: run update-release-notes.sh, or something.

Exactly. Though we could leverage `quarto run` command to write a script in R, python, lua, or typescript instead of `.sh` and the step would be 
- `quarto run update-release-notes.R` for example. 

Like our update for reference pages once per version from script in tools/. That is my idea.

>  Is this something we should do in this PR? Or should we merge this, then figure out the script?

It is not necessary on this PR though one advantage: 
- You know the step you did know, so you could write the script
- You can easily reset the state of the file from change in this PR, run the script and test it gets you exactly what you did manually. 

But it can definitely be done later for next release. So if you prefer that, I would merge this one and open an issue with the idea. I could work on it later for quarto 1.6 release if you want

cwickham:
I want to get the blog post out today anddon't have time for the script right now. So, I'll merge this and open an issue for the script.

Thanks for the feedback!

cwickham:
Opened https://github.com/quarto-dev/quarto-cli/issues/10277

PR #1255, opened 2024-07-10
v1.5 release blog post

cwickham:


mine-cetinkaya-rundel:
The post LGTM! One potential change I might suggest is not placing the placeholder image and text in the margin -- it looks nice there but I was confused as to whether the placement was part of the shortcode.

PR #1266, opened 2024-07-17
Update cross-references-divs.qmd

sgelzenleuchter:
Fix anchor link to section [Listings](https://quarto.org/docs/authoring/cross-references-divs.html#listings).

### DIFF
--- a/docs/authoring/cross-references-divs.qmd
+++ b/docs/authoring/cross-references-divs.qmd
@@ -6,7 +6,7 @@ title: Cross-Reference Div Syntax
 
 ## Overview
 
-Cross-referenceable [figures](#figures), [tables](#tables) and [code listings](#code-listings) are known as *float* cross-references. Floats can appear in the rendered document at locations other than where they are defined, i.e. they float, and usually have captions. 
+Cross-referenceable [figures](#figures), [tables](#tables) and [code listings](#listings) are known as *float* cross-references. Floats can appear in the rendered document at locations other than where they are defined, i.e. they float, and usually have captions. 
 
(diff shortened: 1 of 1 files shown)


docs/authoring/cross-references-divs.qmd
```diff
@@ -6,7 +6,7 @@ title: Cross-Reference Div Syntax
 
 ## Overview
 
-Cross-referenceable [figures](#figures), [tables](#tables) and [code listings](#code-listings) are known as *float* cross-references. Floats can appear in the rendered document at locations other than where they are defined, i.e. they float, and usually have captions.
```
cderv, line 9: This anchor points to https://quarto.org/docs/authoring/cross-references.html#code-listings 

And so the id seems right `#code-listings`- the link at https://quarto.org/docs/authoring/cross-references.html#overview works ok for me.

cderv (commented):
Thanks. 

Not sure about why this update is needed though... 🤔 

Can you give more context ?

docs/authoring/cross-references-divs.qmd
cwickham: The other links (`figures` and `tables`) point to sections on this page (`cross-reference-divs.qmd`) so `code listings` should too. But, the section is actually called `"Listings"` not `"Code Listings"` on this page. I think this change is required. I'll generate preview to check.

cderv: Indeed the preview fixed the link. Thanks for the explanation ! 

And thanks @sgelzenleuchter !

PR #1269, opened 2024-07-18
Update version info and links in licences

cderv:
Following discussion at https://github.com/quarto-dev/quarto-cli/discussions/10319#discussioncomment-10084294, we noticed this page was outdated regarding some version information. 

This PR update it, and add two more entries of tools we ship and use as external tooling 
- Typst for PDF generation
- Juice that we call from Pandoc Lua directly. 

Opening this PR to confirm we want to mentioned them like this. 

Preview: https://deploy-preview-1269.quarto.org/license.html


gordonwoodhull:
Good catch

cwickham (commented):
LGTM

PR #1283, opened 2024-07-27
Mention using imakeidx to build an index

Blake-Madden:
Mention that you can use `imakeidx` to build an index (as an alternative to `makeidx`). This package offers a number of additional features.

Provide an example of some of `imakeidx`'s features.

Fix stray "quart" in paragraph.


cderv:
Thanks

PR #1337, opened 2024-09-17
fix: `docs/projects/scripts.qmd` - fix typo

Gewerd-Strauss:
Found what I assume is a small typo in `docs/projects/scripts.qmd` - at least the previous sentence wasn't really correct 

Sidenote: 

I could not find any specific style guide reg. whitespace in raw markdown-tables, thus I took the freedom to maintain a flush right edge to the table.

~Gw


cderv (approved):
Thanks !

PR #1339, opened 2024-09-17
docs: grid can be set per page

mcanouil:
Since https://github.com/quarto-dev/quarto-cli/pull/10611 `grid` can be set on a page of a website.

- https://github.com/quarto-dev/quarto-cli/discussions/6615
- https://github.com/quarto-dev/quarto-cli/discussions/10818

### DIFF
--- a/docs/output-formats/page-layout.qmd
+++ b/docs/output-formats/page-layout.qmd
@@ -100,9 +100,9 @@ You can control the width of the layout components in HTML documents with YAML o
 ![Wider Sidebar](images/grid-wide-sidebar.png){fig-alt="Screenshot of a Quarto website with altered layout, devoting more space to the sidebar."}
 :::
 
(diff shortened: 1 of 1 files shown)


cderv (commented):
Thanks. 

Other part of this doc should be updated then right ? 
Like https://deploy-preview-1339.quarto.org/docs/output-formats/page-layout.html#grid-customization 
image

Possibly that is the only missing change.

mcanouil:
Well, I thought about changing it here as well, but I thought of it as an example. So to me it make sense to be a bit specific. It also makes sense to keep it general in this part which is kind of a summary.
What do you think?

mcanouil:
I made the change.

cderv:
> Well, I thought about changing it here as well, but I thought of it as an example. So to me it make sense to be a bit specific. It also makes sense to keep it general in this part which is kind of a summary.

Yeah makes sense - you're right about this. I'll revert and merge thanks !

PR #1354, opened 2024-09-26
[gha] Update to node 20

cderv:
closes https://github.com/quarto-dev/quarto-cli/issues/8906

Our custom action is running on node16, we just need to update to node 20 now. Hopefully it is not big change


cderv:
it seems updating the version does work without any more change

PR #1386, opened 2024-10-16
Script to automate release notes

cwickham:
Adds a script to automate the steps of adding release notes for the old release.

Closes https://github.com/quarto-dev/quarto-cli/issues/10277

In addition, we'd need to change this step in the release checklist:

https://github.com/quarto-dev/quarto-cli/blob/a2026c8b7c775665472b73f30b84d8da7a94c178/dev-docs/checklist-make-a-new-quarto-release.md?plain=1#L24

to:
```markdown
  - run `quarto run tools/release-notes.R`
```

### DIFF
--- a/tools/release-notes.R
+++ b/tools/release-notes.R
@@ -1,19 +1,52 @@
 library(fs)
 library(stringr)
 library(glue)
-
-# Version numbers
-old_release <- "v1.5.57"
-new_release_major <- "1.6"
-new_prerelease_major <- "1.7"
-
-major_version <- str_extract(old_release, "(\\d+)\\.(\\d+)")
+library(jsonlite)
+library(gh)
 
 downloads <- path("docs", "download")
 
+# Current versions -------------------------------------------------------
+
+# Based on updated download files
+new_release <- read_json(path(downloads, "_download.json"))$version
+new_prerelease <- read_json(path(downloads, "_prerelease.json"))$version
+
+# Old version from Git history -------------------------------------------
+
+# Need version at two commits ago
+previous_commit <- gh("https://api.github.com/repos/:owner/:repo/commits",
+  owner = "quarto-dev", repo = "quarto-web", 
+  path = "docs/download/_download.json",
+  per_page = 2)
+
+previous_commit_ref <- previous_commit[[2]]$sha
+
+previous_contents <- gh("/repos/{owner}/{repo}/contents/{path}", 
+  owner = "quarto-dev", repo = "quarto-web", 
+  path =  "docs/download/_download.json",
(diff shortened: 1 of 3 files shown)


cwickham:
@cderv Could you take a look at this and see what you think? 

I plan to have the version numbers be provided as command line inputs. I couldn't think of a better way than taking three inputs `old_release`, `new_release_major`, and `new_prerelease_major`, so I'm open to ideas.

cderv:
That looks good ! 

Few ideas that I am throwing here: 

- We can create listing content through a YAML file: https://quarto.org/docs/websites/website-listings.html#yaml-listing-content. This means that for `Add new item to download-older listing in docs/download/index.qmd` we could managing the file programmatically through your script to add a version. Maybe no manual step would be needed in that case. 

- Regarding arguments, I believe we have ways to retrieve version information. 
	- We can fetch information for Github latest release, or even our own API https://quarto.org/docs/download/_download.json or https://quarto.org/docs/download/_prerelease.json
	- From there you can know / compute old, new one and next one. 

I believe with this, you can have a full automated script for this update. Only manual step would be to check git diff and commit. 

What do you think ?

cwickham:
@cderv Ah yeah, the listing idea is good! I'll implement that. Actually, `docs/download/_download-older.yml` exists, we just weren't using it!

If we wait until after `_download.json` and `_prerelease.json` are generated to run this script (which we do according to the checklist), I can use them for the new release version, and new prerelease version. I might need to look into their history to get the old release version.

cwickham:
@cderv OK, I think I've fully automated now. Do you want to take another look?

tools/release-notes.R
```diff
@@ -0,0 +1,98 @@
...
+
+previous_contents <- gh("/repos/{owner}/{repo}/contents/{path}", 
+  owner = "quarto-dev", repo = "quarto-web", 
+  path =  "docs/download/_download.json",
+  ref = previous_commit_ref,
+  .accept = "application/vnd.github.raw+json")
+previous_contents_json <- parse_json(previous_contents$message)
+
+old_release <- previous_contents_json$version
+old_release_date <- previous_contents_json$created
```
cderv, line 33: Isn't this available online in the release website ? 🤔 
at https://quarto.org/docs/download/_download.json 

Could be simpler to not query GH API, but I am sure it works. 

Though I am not sure about the two commits ago... so maybe I am wrong.

cderv: Oh ok. We need to pay attention to when this "two commit back" assumption may not be working anymore then.

cderv (approved):

PR #1421, opened 2024-10-28
Update title-blocks.qmd

melaniewalsh:
I think it's just doi, not doi-title

### DIFF
--- a/docs/authoring/title-blocks.qmd
+++ b/docs/authoring/title-blocks.qmd
@@ -89,7 +89,7 @@ The labels for the metadata included in the title block have default values that
 | `abstract-title`    | Abstract       | `plain`, `default`, `none` |
 | `description-title` | Description    | `plain`, `default`         |
 | `published-title`   | Date Published | `plain`, `default`         |
-| `doi-title`         | DOI            | `plain`, `default`         |
+| `doi`         | DOI            | `plain`, `default`         |
 
 ## Custom Title Pages
 


cderv:
What makes you say that ? 

We do use `doi-title` in quarto CLI
https://github.com/quarto-dev/quarto-cli/blob/464c18c3a4a02df537e1765883eeaf1cafa30f6d/src/resources/filters/modules/authors.lua#L222
https://github.com/quarto-dev/quarto-cli/blob/464c18c3a4a02df537e1765883eeaf1cafa30f6d/src/resources/filters/modules/authors.lua#L890-L892

`doi-title` is the way to change the label used which is `Doi` by default. `doi` meta is for the DOI reference itself AFAIK

melaniewalsh:
@cderv In my version of Quarto (1.5.47), `doi` in the YAML adds a DOI to the title block but `doi-title` does nothing.

For example, this is `doi`:
Screenshot 2024-11-05 at 9 36 50 AM

This is `doi-title`:
Screenshot 2024-11-05 at 9 37 46 AM

You can see it in our repo, as well: https://github.com/Post45-Data-Collective/website/blob/29d6cd03b8197232d0196dd3a9213de00ae8240d/posts/the-index-of-major-literary-prizes-in-the-us/index.qmd#L8

melaniewalsh:
@cderv Wait ok, I understand now that`doi-title` is changing the label for the DOI in the title block. I still think this is confusing in the documentation though because I don't think it specifies elsewhere how to just add a DOI to the title block?

cderv:
> because I don't think it specifies elsewhere how to just add a DOI to the title block?

This is documented only in the reference page: https://quarto.org/docs/reference/formats/html.html#title-author

We could indeed add it in the guide part of the doc - just need to find the right place for it. 

possibly by adding a new table in https://quarto.org/docs/authoring/title-blocks.html about allowed fields

melaniewalsh:
I think a new table on the title blocks page would be great. Thanks for pointing out where it's documented in the reference page. I think when you do a global search on the site, the guide pages come up first, so having it documented in the guide would help (and maybe linking to the reference would help, too). Thanks!

cderv:
I opened the idea to track this
- https://github.com/quarto-dev/quarto-cli/issues/11322

Thanks for the feedback !

PR #1423, opened 2024-10-28
Brand reference

cwickham:
First attempt: only lists options at first two levels. 

Will close https://github.com/quarto-dev/quarto-cli/issues/11178


gordonwoodhull:
I've added one level of recursion for the typography options.

There are a few places where we could go deeper:
- [ ] We could e.g. also show `brand-font-weight`. 
- [ ] Logo has deeper structure like `brand.logo.medium.dark` (but it's hard to see what to do about it).
- [ ] [Font resource definitions](https://posit-dev.github.io/brand-yml/brand/typography.html#comprehensive-example-with-font-definitions).

Happy to iterate on this.

cwickham:
This looks good! Of your other options, I'd like to see the "Font resource definitions". 

I think this is pretty close to the level of detail that we have for some other reference sections. Examples of e.g. `brand-font-weight` can be given in the Guide. Same with `logo` details.

PR #1436, opened 2024-11-04
bump to latest Rstudio IDE version

cderv:
closes quarto-dev/quarto-cli#11298

@cwickham do you want to still keep track on version number and update this variable regularly ? 

or should we just say "latest" with a link to https://posit.co/download/rstudio-desktop/ and stop showing the version. 

We don't update our doc at each release on it... 

Otherwise, we could come up with a pre-render script that would query the API at https://posit.co/wp-content/uploads/downloads.json

### DIFF
--- a/_variables.yml
+++ b/_variables.yml
@@ -1,3 +1,2 @@
 rstudio:
   min_version: "v2022.07"
-  current_release: "v2024.09"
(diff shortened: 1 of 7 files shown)


cwickham:
You are right, we could probably just say "latest". I've push a commit that makes that change.

PR #1437, opened 2024-11-04
11300 - clarify that section title are not scrollable

GuillaumeDehaene:
Related https://github.com/quarto-dev/quarto-cli/issues/11300

Update documentation to clarify:

- that section title slides are not scrollable by default,
- how to modify them to make them scrollable.

### DIFF
--- a/docs/presentations/revealjs/index.qmd
+++ b/docs/presentations/revealjs/index.qmd
@@ -55,6 +55,18 @@ format:
 ---
 ```
 
+Please note that section title slides are centered by default which prevents them from being scrollable.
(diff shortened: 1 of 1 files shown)


cderv:
closes https://github.com/quarto-dev/quarto-cli/issues/11300

@cwickham this looks to me. This is indeed true and worth knowing. 

I let you decide if this is ok shown as is, or if it should be in a callout

cwickham:
Looks good to me. I don't think it needs to be a callout. Only one comment ☝🏻.

GuillaumeDehaene:
Accepted the change

cderv (approved):

PR #1444, opened 2024-11-05
feat: add inline-svg filter

adamblake:
As requested by @RoyiAvital (https://github.com/coursekata/quarto-inline-svg/issues/1)


cderv:
Thanks. 

Curious about something: This extension does the same thing as pandoc's `embed-resources` for svg right ?
Is the use case to have inline svg for image, without having all the resources to be embedded ? 

That is a nice feature IMO ! 

Also, I saw you have a `base64` version in Lua. Just to share that we maintain a version of it in Quarto API ICYDK: https://quarto.org/docs/extensions/lua-api.html#base64-encoding

adamblake:
@cderv first I just want to make sure I understand what you're saying. This is what I think you're saying: if you use `embed-resources`, that will wrap everything up into your output files, but if you just want to embed the SVG images, this filter is the tool you want.

If my understanding is correct, I think I should change the description of this to add something to that effect!

Mainly my use-case or rather thinking was that the generated images from code on each page are not re-usable resources. As such I just wanted to simplify the output directory and put everything in a single page. This makes it easier to do something like upload the source to an S3 without having to search for related files or whatever.

adamblake:
Also, since this was my first time writing a Lua filter (and last time I touched it), is this how I would update the filter to use the Quarto base64? I have to admit I don't understand how packaging works in Lua, but it seems like of like JS where the Quarto packages have already been loaded for use by extensions...?

**Before**:
```lua
-- Find references to generated SVGs and replace with SVGs.
local formats = { "html*", "markdown*", "commonmark*", "gfm", "markua" }
if match_any(FORMAT, formats) then
  function Image(elem)
    if file_exists(elem.src) and ends_with(elem.src, ".svg") then
      local svg = read_svg(elem.src)
      elem.src = "data:image/svg+xml;base64," .. base64(svg)
      return { elem }
    end
  end
end
```

**After**:
```lua
```lua
-- Find references to generated SVGs and replace with SVGs.
local formats = { "html*", "markdown*", "commonmark*", "gfm", "markua" }
if match_any(FORMAT, formats) then
  function Image(elem)
    if file_exists(elem.src) and ends_with(elem.src, ".svg") then
      local svg = read_svg(elem.src)
      elem.src = "data:image/svg+xml;base64," .. quarto.base64.encode(sig)
      return { elem }
    end
  end
end
```

cderv:
> if you use embed-resources, that will wrap everything up into your output files, but if you just want to embed the SVG images, this filter is the tool you want.

Yes exactly. When using `embed-resources`, any svg file will be inlined in the document. You can try on your example, by removing the filter configuration. 

> Mainly my use-case or rather thinking was that the generated images from code on each page are not re-usable resources. 

This is definitely a good idea, that we could have leveraged into **knitr** directly. Your extension is even in better as it apply to all engine, not just knitr. 

> but it seems like of like JS where the Quarto packages have already been loaded for use by extensions...

Yes any API we describe in https://quarto.org/docs/extensions/lua-api.html is available to be used in filter **in Quarto context**. There is no need to load anything. `quarto` object will be defined, and using `quarto.base64.encode` should work directly

adamblake:
@cderv thanks for the feedback and clarifications! I updated the description here and in my README to note the similarity and difference from `embed-resources` and I pushed a new version that uses the Quarto namespace base64 implementation

PR #1452, opened 2024-11-13
Move giscus comments to quarto-cli repo discussion board

cderv:
This should be enough to move discussion from quarto-web to quarto-cli

I created the category there 
https://github.com/quarto-dev/quarto-cli/discussions/categories/blog

I just need to move every post from here to there so that current comments will work. 

I'll try to make a script for that so that we can sync up the merge of this PR with the move. Maybe do it before merging. 

I'll update when done


cderv:
Marking as draft until I move all the discussions otherwise, it will mess up and create duplicate.

PR #1461, opened 2024-11-15
Brand edits

cwickham:


docs/authoring/brand.qmd
```diff
@@ -434,19 +442,62 @@ You can't currently access `typography`, `meta`, or `defaults` values using the
...
     blue: "#ddeaf1"
 ```
 
-This should work:
+Then the variable `$brand-blue` is will be set to `#ddeaf1` in the `defaults` layer of  [Quarto's SCSS layering system](/docs/output-formats/html-themes-more.html#bootstrap-bootswatch-layering). 
+You can add a custom SCSS file, `styles.scss`, in the usual way:
+```{.yaml filename="_quarto.yml"}
+format:
+  html:
+    theme: [styles.scss]
```
cwickham, line 459: @cscheid Does all of this apply to `revealjs` too? Do we expect people to just do:
```suggestion
theme: [styles.scss]
```

docs/authoring/brand.qmd
```diff
@@ -434,19 +442,62 @@ You can't currently access `typography`, `meta`, or `defaults` values using the
...
 ```{.yaml filename="_brand.yml"}
 color:
   palette:
     blue: "#ddeaf1"
 ```
 
-This should work:
+Then the variable `$brand-blue` is will be set to `#ddeaf1` in the `defaults` layer of  [Quarto's SCSS layering system](/docs/output-formats/html-themes-more.html#bootstrap-bootswatch-layering). 
+You can add a custom SCSS file, `styles.scss`, in the usual way:
+```{.yaml filename="_quarto.yml"}
```
cwickham, line 456: @cscheid  Do you think we should do this in `_quarto.yml` or should I do it in `_brand.yml` under `defaults: quarto: formats:`?
That raises another question, can I set things directly under `defaults: quarto` and they'll apply to all formats? E.g.
```
defaults:
  quarto:
    toc: true
```

docs/authoring/brand.qmd
```diff
@@ -434,19 +442,62 @@ You can't currently access `typography`, `meta`, or `defaults` values using the
...
+```
+
+When you specify SCSS files without an explicit `brand` element, it is equivalant to listing them after `brand`.
+For instance, the above `theme` is equivalant to:
+
+```{.yaml filename="_quarto.yml"}
+format:
+  html:
+    theme: [brand, styles.scss]
+```
```
cwickham, line 480: @cscheid Is this correct? Not specifying `brand` puts it first.

docs/authoring/brand.qmd
```diff
@@ -434,19 +442,62 @@ You can't currently access `typography`, `meta`, or `defaults` values using the
 
 ### In SCSS
 
-If `_brand.yml` defines `color: blue`:
+The colors defined in `palette` are set as SCSS variables of the form `brand-COLOR_NAME`.
```
cwickham, line 445: @cscheid Are there other things in brand exposed in SCSS we should mention?

PR #1471, opened 2024-11-21
Environment variable is set in Quarto 1.6 only

cderv:
I believe this is wrong version right ?


cderv:
Oh I did not understood it that way. Ok makes sense.

PR #1545, opened 2025-01-17
feat: Add blog post for GitHub Pages

mcanouil:
Introduce a new blog post along with a featured image for publication on GitHub Pages.

Originally published on [mickael.canouil.fr](https://mickael.canouil.fr/posts/2024-12-30-quarto-github-pages/).


cwickham (commented):
I think there is some useful content in here but it's a bit buried.

The first two scenarios, Manual, and Automated - Quarto Publish GitHub Action, are what we describe in the docs. I think you could minimise their content and stick to "these work well in ___ cases". Then, really highlight the cases they don't work and use that to introduce the two scenarios in "Custom GitHub Actions Workflow". 

I also think it would help if you called out the "new" steps in the workflows (maybe with code annotation). So, someone can easily see what is different, and also understand what each step is doing (especially those shell commands).

mcanouil:
I almost forgot about this.
Based on your comments, It feels like it does not add much value (compared to the original blog post) and quite an amount of work.
I'm not sure that I can spend the required time, so I would close this unless someone wants to do the final editing because there is still value in it and on its own.

PR #1549, opened 2025-01-20
Give an example with tbl-pos="t" that also needs layout-ncol="1" to w…

fuhrmanator:
…ork in PDF

See the comment at https://github.com/quarto-dev/quarto-cli/discussions/4849#discussioncomment-11622517

I'm suggesting an example be added to the documentation, as it took me a long time to figure this out!

### DIFF
--- a/docs/authoring/cross-references-divs.qmd
+++ b/docs/authoring/cross-references-divs.qmd
@@ -136,6 +136,23 @@ In the above example, you can reference the figure with `@fig-line-plot`, but no
 
 {{< include _cross-reference-divs-table.qmd >}}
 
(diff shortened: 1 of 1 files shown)


fuhrmanator:
looks good to me at https://deploy-preview-1549.quarto.org/docs/authoring/cross-references-divs.html#tables

image

cderv:
Thanks, this looks good indeed. 

Re-reading your example in preview raises one question:  Why `layout-ncol=1` is needed ? This looks like a problem that should be solved, and not just documented officially. 🤔 

Did we look into that in an issue and not just the comment linked ?

PR #1556, opened 2025-01-23
docs: move link outside of footnote

mcanouil:
Currently the link to the full documentation on inline code is "hidden" inside the footnote.
It would be better for it to be clearly visible.
The footnote is mostly there to say the `knitr` syntax is supported.

### DIFF
--- a/docs/get-started/computations/rstudio.qmd
+++ b/docs/get-started/computations/rstudio.qmd
@@ -232,7 +232,7 @@ See the documentation on [Article Layout](/docs/authoring/article-layout.qmd) to
 
 ## Inline Code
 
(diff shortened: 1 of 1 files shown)


cderv (commented):
Thanks. 

It would indeed be good to not link only from the footnote. 

Though I find the "Read more in inline code" not well place. 

I think I would add something at the end of the paragraph like in next "Caching" part 

> You can find out more about caching in Quarto documents in the [Cache](https://deploy-preview-1556.quarto.org/docs/projects/code-execution.html#cache) documentation.

Also I would keep the link in the footnote too. 

What do you think ?

cderv (approved):

PR #1573, opened 2025-03-05
Expand docs for `julia` engine

MichaelHatherly:
As requested in https://github.com/quarto-dev/quarto-cli/pull/12121#issuecomment-2672641459. Adds some additional docs to the section covering the `julia` engine that highlights some or the main features it currently has.

### DIFF
--- a/docs/computations/julia.qmd
+++ b/docs/computations/julia.qmd
@@ -318,6 +318,148 @@ The currently available options are:
 - `exeflags`: An array of strings which are appended to the `julia` command that starts the worker process. For example, a notebook is run with `--project=@.` by default (the environment in the directory where the notebook is stored) but this could be overridden by setting `exeflags: ["--project=/some/directory/"]`.
 - `env`: An array of strings where each string specifies one environment variable that is passed to the worker process. For example, `env: ["SOMEVAR=SOMEVALUE"]`.
 
+### `juliaup` integration
+
+[`juliaup`](https://github.com/JuliaLang/juliaup) is the recommended way to
+install and manage Julia versions. The `julia` engine supports using
+`juliaup`'s `+` channel specifier to select the Julia version that a notebook
+uses. This allow users to run several notebooks concurrently with different
+Julia versions on the same machine without the need for customizing their
+`PATH` in any way.
+
+To use this feature, ensure that you have used `juliaup` to install the channels
+that you wish to use in your notebooks. Then add the channel specifier to the notebook
+frontmatter in the `julia.exeflags` option:
+
+````markdown
+---
+title: "A Julia 1.11 notebook"
+engine: julia
+julia:
+  exeflags: ["+1.11"]
+---
+
+```{{julia}}
+VERSION
+```
+````
+
+`QuartoNotebookRunner` currently supports running Julia versions from 1.6
+onwards. Support for earlier versions is not planned.
+
+### Revise.jl integration
+
+[Revise](https://github.com/timholy/Revise.jl) allows for automatically
+updating function definitions in Julia sessions. It is an essential tool in the
+Julia ecosystem for any user that wishes to develop their own packages or large
+projects. The `julia` engine supports using `Revise` within notebook processes
+by simply importing the package into a cell at the start of a notebook. Once
+imported `Revise` will automatically update functions imported from locally
+developed packages in the same way as it behaves in a Julia REPL.
+
+Ensure that `Revise` is installed in the project environment that the notebook
+is using, since the global environment is not included in the load path
+provided to Julia, unlike the behaviour of a Julia REPL session.
+
+### Caching
+
+The engine has built-in support for caching notebook results. This feature is disabled by
+default but can be enabled by setting `execute.cache` option to `true` in a notebook's
+frontmatter:
+
+````markdown
+---
+title: "A caching example"
+engine: julia
+execute:
+  cache: true
+---
+
+```{{julia}}
+rand()
+```
+````
+
+Notebook caches are invalidated based on the following criteria:
+
+- Using a different Julia version to run the notebook.
+- The `Manifest.toml` of the environment the notebook is run in. Adding, removing, or changing package versions will invalidate the cache.
+- Changing any contents in the notebook's frontmatter.
+- Changing any contents of any executable Julia cells, including cell options.
+
+Changes that do not invalidate a cache:
+
+- Editing markdown content outside of executable Julia cells.
+
+Caches are saved to file in a `.cache` directory alongside the notebook file.
+This directory is safe to remove if you want to invalidate all caches. The
+contents of the individual cache files is an internal implementation detail and
+should not be relied upon.
+
+These caches are safe to save in CI via such tools as GitHub Actions
+`action/cache` to help improve the render time of long-running notebooks that
+do not often change.
+
+### R and Python support
+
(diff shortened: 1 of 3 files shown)


MichaelHatherly:
@jkrumbiegel any other features you can think of that would be worth mentioning here?

docs/computations/julia.qmd
```diff
@@ -318,6 +318,148 @@ The currently available options are:
...
+### `juliaup` integration
+
+[`juliaup`](https://github.com/JuliaLang/juliaup) is the recommended way to
+install and manage Julia versions. The `julia` engine supports using
+`juliaup`'s `+` channel specifier to select the Julia version that a notebook
+uses. This allow users to run several notebooks concurrently with different
+Julia versions on the same machine without the need for customizing their
+`PATH` in any way.
+
+To use this feature ensure that you have used `juliaup` to install the channels
```
jkrumbiegel, line 330: "feature, ensure" I think

docs/computations/julia.qmd
```diff
@@ -318,6 +318,148 @@ The currently available options are:
...
+that you wish to use in your notebooks. The add the channel specifier to the notebook
```
jkrumbiegel, line 331: "Then add"

jkrumbiegel (commented):
looks good to me otherwise, I can't think of anything big that's missing

docs/computations/julia.qmd
```diff
@@ -318,6 +318,148 @@ The currently available options are:
...
+projects. The `julia` engine supports using `Revise` within notebook processes
+by simply importing the package into a cell at the start of a notebook. Once
+imported `Revise` will automatically update functions imported from locally
+developed packages in the same way as it behaves in a Julia REPL.
+
+Ensure that `Revise` is installed in the project environment that the notebook
+is using, since the global environment is not included in the load path
+provided to Julia, unlike the behaviour of a Julia REPL session.
+
+### Caching
```
cderv, line 364: Now that julia engine support caching, a link to this part from this part could be useful: https://quarto.org/docs/projects/code-execution.html#cache

BTW do `--cache-refresh` works with Julia engine ?

docs/computations/julia.qmd
```diff
@@ -318,6 +318,148 @@ The currently available options are:
 - `exeflags`: An array of strings which are appended to the `julia` command that starts the worker process. For example, a notebook is run with `--project=@.` by default (the environment in the directory where the notebook is stored) but this could be overridden by setting `exeflags: ["--project=/some/directory/"]`.
 - `env`: An array of strings where each string specifies one environment variable that is passed to the worker process. For example, `env: ["SOMEVAR=SOMEVALUE"]`.
 
+### `juliaup` integration
```
cderv, line 321: This is cool ! Should `juliaup` be mentioned in the installation section: https://quarto.org/docs/computations/julia.html#installation-1 ?

cderv (approved):

docs/computations/julia.qmd
MichaelHatherly: > BTW do --cache-refresh works with Julia engine ?

Probably not currently. Do you happen to know if CLI flags get forwarded to the `julia` process along with all the frontmatter metadata from the notebook? If it does then we could probably support that behaviour.

docs/computations/julia.qmd
MichaelHatherly: Probably best if it was.

docs/computations/julia.qmd
cderv: I have quickly look 

`--cache-refresh` does set internal variable
https://github.com/quarto-dev/quarto-cli/blob/0502c32d72bbc90cdd4736135bd7f645e773ff37/src/command/render/flags.ts#L272-L275

that is use to modify `execute.cache` 
https://github.com/quarto-dev/quarto-cli/blob/0502c32d72bbc90cdd4736135bd7f645e773ff37/src/command/render/render-contexts.ts#L190-L193

so basically this is same as
```
execute:
  cache: refresh
```

So you have access to it already same way as Jupyter 
https://github.com/quarto-dev/quarto-cli/blob/0502c32d72bbc90cdd4736135bd7f645e773ff37/src/resources/jupyter/notebook.py#L116-L139

Does it help ?

MichaelHatherly: Yes, that does help, thanks.

cwickham:
I think a "highlight" is the right way to expose these changes. I've added one.

docs/computations/julia.qmd
```diff
@@ -12,6 +12,13 @@ Quarto supports executable Julia code blocks within markdown. This allows you to
 
 Quarto has two available engines for executing Julia code. The older one is using the [IJulia](https://github.com/JuliaLang/IJulia.jl) Jupyter kernel and depends on Python to run. The newer engine is using the [QuartoNotebookRunner.jl](https://github.com/PumasAI/QuartoNotebookRunner.jl/) package to render notebooks and does not have any additional dependencies beyond a Julia installation.
 
+To use either of these engines will require that you manually install Julia if
+you have not done so already. You can download it from
```
cwickham, line 16:
```suggestion
Using either of these engines will require manually installing Julia if
you have not done so already. You can download it from
```

docs/computations/julia.qmd
```diff
@@ -12,6 +12,13 @@ Quarto supports executable Julia code blocks within markdown. This allows you to
...
+https://julialang.org/downloads/. There are two options for installation: the
+Juliaup installation manager, and the "manual" installation approach. Unless
+you know that you need to use the "manual" approach you should prefer using
+Juliaup since it allows you to manage multiple Julia versions on your system.
```
cwickham, line 20:
```suggestion
https://julialang.org/downloads/. There are two options for installation: the
Juliaup installation manager, or the "manual" installation approach. Unless
you know that you need to use the "manual" approach, use
Juliaup since it allows you to manage multiple Julia versions on your system.
```

docs/computations/julia.qmd
```diff
@@ -318,6 +325,148 @@ The currently available options are:
...
+developed packages in the same way as it behaves in a Julia REPL.
+
+Ensure that `Revise` is installed in the project environment that the notebook
+is using, since the global environment is not included in the load path
+provided to Julia, unlike the behaviour of a Julia REPL session.
+
+### Caching {#caching-julia}
+
+The engine has built-in support for caching notebook results. This feature is disabled by
+default but can be enabled by setting `execute.cache` option to `true` in a notebook's
```
cwickham, line 374:
```suggestion
default but can be enabled by setting the `execute.cache` option to `true` in a notebook's
```

docs/computations/julia.qmd
```diff
@@ -318,6 +325,148 @@ The currently available options are:
...
+
+```{{julia}}
+rand()
+```
+````
+
+Notebook caches are invalidated based on the following criteria:
+
+- Using a different Julia version to run the notebook.
+- The `Manifest.toml` of the environment the notebook is run in. Adding, removing, or changing package versions will invalidate the cache.
```
cwickham, line 393: The first sentence is a fragment...maybe?
```suggestion
- Changes to the `Manifest.toml` of the environment the notebook is run in. Adding, removing, or changing package versions will invalidate the cache.
```

cwickham (approved):
Just a few wording suggestions.

MichaelHatherly:
Anything else needed in this PR or is this able to be merged?

cderv (approved):

PR #1583, opened 2025-03-11
Add custom domain set up to Github Pages

cwickham:
* Adds a ["Custom Domain"](https://deploy-preview-1583.quarto.org/docs/publishing/github-pages.html#custom-domain) section to Github Pages
* Spells out files Quarto knows to copy to output directory in [Site Resources](https://deploy-preview-1583.quarto.org/docs/websites/website-tools.html#site-resources)

Closes https://github.com/quarto-dev/quarto-cli/issues/4941

### DIFF
--- a/docs/publishing/github-pages.qmd
+++ b/docs/publishing/github-pages.qmd
@@ -311,7 +311,7 @@ blog.example.com
 ```
 ::::
 
-Quarto recognizes the `CNAME` file and will copy it your site output directory when you render.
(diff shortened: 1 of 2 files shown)


cwickham:
@mcanouil , I'm not convinced I followed the whole thread in the related issues and discussion. Does this cover it? It seems like the same instructions work regardless of method (`docs/`, `quarto publish gh-pages`, or GitHub Action). Is that right?

mcanouil:
I think that covers the current resources detection from Quarto (via `quarto render`).
So it should cover manual deployment and deployment via the publish command.

cderv (approved):
Looks good to me! Thanks a lot !!

PR #1604, opened 2025-04-01
Mention `caption-location` on guide page

cwickham:
Closes https://github.com/quarto-dev/quarto-cli/issues/11831


cwickham:
@cderv Could you check the changes to `_freeze/` are expected here. I think they come from rendering: https://deploy-preview-1603.quarto.org/docs/presentations/revealjs/index.html I used v1.7.20 locally.

PR #1612, opened 2025-04-03
run post-render script only on project render

cderv:
This will prevent the script to run on incremental render and preview


cderv:
@cwickham This is assuming we want to render the post-render script only for `quarto render`. 

If we want also for preview or incremental render, then current behavior is ok.

cwickham (approved):
@cderv Yes, this seems like a good idea. We only really need to run that script on CI.

PR #1614, opened 2025-04-03
docs: dark brand

gordonwoodhull:
Hi @cwickham, here is the first of my doc PRs for Quarto 1.7, covering Dark Brand.

I'll follow up with more commits covering the changes for Typst, and a separate PR for `renderings`. 

(I will also need to check that my schema changes have made it into the reference documentation, hmm.)

Please let me know what you think - glad to edit, move content, etc.

Thanks!


cwickham (commented):
This all looks great. A few places that could use a bit more clarification.

I'll open a PR for the updates to the reference so you can take a look.

docs/output-formats/html-themes.qmd
```diff
@@ -101,10 +101,12 @@ Setting the above themes in your `_quarto.yml` results in both a dark and light
 
 When providing both a dark and light mode for your html output, Quarto will automatically create a toggle to allow your reader to select the desired dark or light appearance. The toggle will automatically appear in the top right corner of your html output. When possible, the toggle will use browser local storage to maintain the user's preference across sessions.
 
-The first appearance (light or dark) elements in the theme to determine the default appearance for your html output. For example, since the `light` option appears first in the above example, a reader will see the light appearance by default.
+To honor the user's operating system or browser preference for light or dark mode, specify `respect-user-color-scheme: true`. Otherwise, the order of light and dark elements in the theme or brand will determine the default appearance for your html output. For example, since the `light` option appears first in the above example, a reader will see the light appearance by default, if `respect-user-color-scheme` is not enabled.
```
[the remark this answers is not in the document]
gordonwoodhull: Good idea. It's a new HTML document option. Will do.

docs/output-formats/html-themes.qmd
```diff
@@ -101,10 +101,12 @@ Setting the above themes in your `_quarto.yml` results in both a dark and light
...
 
 Quarto will automatically select the appropriate light or dark version of the text highlighter that you have specified when possible. For more information, see [Code Highlighting](html-code.qmd#highlighting).
 
+As of Quarto 1.7, `respect-user-color-scheme` applies on page load, and the page won't change if dark mode is changed dynamically. It also requires JavaScript support: users with JavaScript disabled will see the author-preferred (first) brand or theme.
```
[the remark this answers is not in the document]
gordonwoodhull: It's possible for a page to change when the user changes their OS preference. Or the user may have their OS set to use light mode during the day and dark mode at night.

We don't support that yet.

I'll think about a better way to say this. The reference documentation all shows up — nice system! And it points out I have do a round of copy editing in the schema descriptions, including this option.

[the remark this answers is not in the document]
gordonwoodhull: I've removed this language because I'm filing a PR to implement changing the page when the OS option changes.
https://github.com/quarto-dev/quarto-cli/pull/12465

docs/output-formats/html-themes.qmd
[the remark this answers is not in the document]
gordonwoodhull: Adding the example meant invalidating the reference to "above example" in the next sentence, so I changed it to "first example". Is this clear?

gordonwoodhull:
I've added documentation for Typst `brand-mode` and resolved the other issues.

@cwickham, please give this another read for clarity.

docs/authoring/brand.qmd
```diff
@@ -214,6 +255,63 @@ The **brand.yml** specification allows you to specify a `light` and `dark` versi
 
 ::: 
 
+#### Document logo customization
+
+The document `logo` option allows extra customization of the logo.
```
cwickham, line 260:
```suggestion
You can further customize the logo at the document-level with the `logo` option.
```

docs/authoring/brand.qmd
```diff
@@ -214,6 +255,63 @@ The **brand.yml** specification allows you to specify a `light` and `dark` versi
...
+
+To choose a brand logo resource:
```
cwickham, line 262:
```suggestion
For example, you can specify a particular brand logo resource:
```

docs/authoring/brand.qmd
```diff
@@ -214,6 +255,63 @@ The **brand.yml** specification allows you to specify a `light` and `dark` versi
...
+```{.yaml filename="document.qmd"}
+---
+logo: large
+---
+```
+
+To choose a brand logo resource and change the alt text:
```
cwickham, line 269:
```suggestion
Or, specify a brand logo and change the alt text:
```

docs/authoring/brand.qmd
```diff
@@ -214,6 +255,63 @@ The **brand.yml** specification allows you to specify a `light` and `dark` versi
...
+```{.yaml filename="document.qmd"}
+---
+logo:
+  width: 1in
+  location: right-top
+  padding-right: 0.5in
+  padding-top: 0.25in
+  alt: Alternate alternate text
+---
+```
```
cwickham, line 298: I think we should specifically target `typst` in the YAML of this example. This?
````suggestion
```{.yaml filename="document.qmd"}
---
format: typst
logo:
  width: 1in
  location: right-top
  padding-right: 0.5in
  padding-top: 0.25in
  alt: Alternate alternate text
---
```
````
Or maybe nest `logo` under `typst` to really emphasize these are `typst` only options.

**Review by cwickham — COMMENTED** [review]
Looks good! A few minor suggestions.

In terms of `renderings`:
* We definitely need some examples. Can we do screenshots for now? Not sure how big a job it is to implement dark mode on quarto.org.  
* Where were you thinking of putting the examples? Maybe `docs/computations/execution-options.html#figure-options`? 
* I think you might have mentioned you can construct light/dark variations of (e.g.) screenshots by constructing divs with the right classes. Is this something we want to document?

It's starting to feel like the dark mode details are spread all over the place and we might want a single page that describes implenting dark mode on a website: setting themes, brand and renderings.

File: docs/authoring/brand.qmd (hunk 6 above)
Comment (gordonwoodhull) [reply docs/authoring/brand.qmd:298]: I like nesting it under `typst`! I had to test it to make sure it works. I'll do that.

**Comment by gordonwoodhull** [comment]
#### Light/dark renderings example

A problem with adding an example today is that we don't have the helper libraries... I'm trying to put the release and docs to bed so that I can get back to the user side. 

(This might also include syntactic sugar to avoid duplicate code, like the custom print function you came up with.)

So a minimal ggplot example is about 56 lines, but 34 lines of this (the first code cell) is the stuff that is supposed to in the helper library:

````
---
title: "knitr dark mode - ggplot"
brand:
  light: united-brand.yml
  dark: slate-brand.yml
execute:
  echo: false
  warning: false
---

```{r}
#| echo: false
#| warning: false
library(ggplot2)

ggplot_theme <- function(bgcolor, fgcolor) {
  theme_minimal(base_size = 11) %+%
    theme(
      panel.border = element_blank(),
      panel.grid.major.y = element_blank(),
      panel.grid.minor.y = element_blank(),
      panel.grid.major.x = element_blank(),
      panel.grid.minor.x = element_blank(),
      text = element_text(colour = fgcolor),
      axis.text = element_text(colour = fgcolor),
      rect = element_rect(colour = bgcolor, fill = bgcolor),
      plot.background = element_rect(fill = bgcolor, colour = NA),
      axis.line = element_line(colour = fgcolor),
      axis.ticks = element_line(colour = fgcolor)
    )
}

brand_ggplot <- function(brand_yml) {
  brand <- yaml::yaml.load_file(brand_yml)
  ggplot_theme(brand$color$background, brand$color$foreground)
}

united_theme <- brand_ggplot("united-brand.yml")
slate_theme <- brand_ggplot("slate-brand.yml")

colour_scale <- scale_colour_manual(values = c("darkorange", "purple", "cyan4"))
```

### no crossref, no caption

```{r}
#| renderings: [light, dark]
ggplot(mtcars, aes(mpg, wt)) +
  geom_point(aes(colour = factor(cyl))) +
  united_theme +
  colour_scale
ggplot(mtcars, aes(mpg, wt)) +
  geom_point(aes(colour = factor(cyl))) +
  slate_theme +
  colour_scale
```
````

I think it would be okay at 12 lines, and I agree it fits in `docs/computations/execution-options.html#figure-options`.

#### non-computational light and dark

We can add this but I'm not sure where it goes. It's not really brand or theme but just the way our CSS works.

A minimal example without any resources would be

````
---
title: "light and dark with divs"
brand:
  light: united-brand.yml
  dark: slate-brand.yml
---

::: {.light-content}
This text will be shown in light mode.
:::

::: {.dark-content}
This text will be shown in dark mode.
:::
````

This example is self-explanatory but kind of useless. Would it be better with images?

**Comment by cwickham** [comment]
I think the crucial bit for people to understand is that they need to generate as many outputs as `renderings`. So, while the ggplot examples are probably the most realistic, we could get away with something simpler to just convey the idea, e.g: 
````
---
title: "Dark mode"
format: 
  html:
    theme:
      light: flatly
      dark: darkly
---


```{r}
#| renderings: [light, dark]
par(bg = "#FFFFFF", fg = "#000000")
plot(1:10) # Shown in `light` mode

par(bg = "#000000", fg = "#FFFFFF", col.axis = "#FFFFFF")
plot(1:10) # Shown in `dark` mode
```

````


Re: non-computational light and dark. Yeah, I see the problem. The cross-format version is pretty ugly...We could do the HTML-only version on a "Dark mode" page. 

I say we merge this (I'll leave whether you want a minimal `renderings` example in this PR or not). 
I think a "Dark mode page" would be in addition to what we have here (maybe just compiling these sections via `include`) and I'm happy to take a stab at that once this is merged.

gordonwoodhull:
Great! You're right, that's exactly what people need to know.

I've added your simple example and will verify and merge.

Thanks!

PR #1615, opened 2025-04-03
Update reference pages

cwickham:
Updating reference pages based on `quarto-cli` `main`.


cwickham:
@gordonwoodhull, take a look to see if changes here capture what you expect for the dark mode updates.

gordonwoodhull:
Thanks @cwickham!

PR #1620, opened 2025-04-06
Update Lua filter which translates CSS properties to Typst link

ryanzomorrodi:
It seems the file name was changed.


cderv (approved):
Thank you !

PR #1652, opened 2025-05-04
Add context and new example to redirects doc

mfisher87:
Hello all, I was reading the redirects doc and felt some additional context could be added to help users better understand why they need to write their config the way they do. I also felt that it would be valuable to have an explicit example for each of the use cases mentioned in the doc, so I added an example for re-organizing directory hierarchy.


mfisher87:
Happy to help, thanks for offering addition clarity in your edits :D

PR #1659, opened 2025-05-07
Feature/cli autogen

cwickham:


cwickham:
> I wonder if we could find a better title for the sidebar than "CLI Commands". "Command-line Reference" is a mouthful, but I don't know if folks know the "CLI" initialism, and "CLI Commands" reads a bit like "ATM Machine" to me :)

Yeah, good catch. Based on the breadcrumbs maybe just something like: `quarto`
Screenshot 2025-05-08 at 1 31 01 PM

cwickham:
And we can use more verbose terms in `docs/cli/index.qmd` and   `docs/reference/index.qmd`

cwickham:
@cscheid Could cli-info also capture whether the command is hidden() ([example](https://github.com/quarto-dev/quarto-cli/blob/fa8476e59848ba76d7b64d1d582d98ae0241f7bc/src/command/editor-support/cmd.ts#L15))? That way the auto-gen reference could display only those shown in `quarto --help`

PR #1665, opened 2025-05-14
Add toggle

coatless:


coatless:
@cderv I thought you folks had a GH action to reorder the listings alphabetically?

cderv:
Nope, we don't. I created a script to reorder at once, but I did not create a GH action that would automatically reorder.	I could have - but you know... time and priority 😅 

It is easy enough to do manually. 😉

PR #1677, opened 2025-06-01
docs: Improve custom listing using raw HTML

mcanouil:
Enhance clarity and formatting in the custom listing documentation, ensuring better understanding and usability for users creating EJS templates.

### DIFF
--- a/docs/websites/website-listings-custom.qmd
+++ b/docs/websites/website-listings-custom.qmd
@@ -8,8 +8,7 @@ In addition to the 3 built in types of listings, you can also build a completely
 
 ## Listing Templates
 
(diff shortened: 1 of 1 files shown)


cderv (commented):
I think I would revisit the order and explain, for the start, that EJS templates are processed as a markdown document, and explain that it is best to use Raw HTML to
- avoid performance issue when processing part that are meant to be HTML with no Markdown
- avoid escaping problem for those part too

Then show an example for a simple template having only HTML, and then your example with mixing Markdown and HTML. 

Internally, we use `.ejs.md` extension, not sure why we don't use that too for our recommendation. This makes it kind of clear

Anyhow, just sharing though. I'll let @cwickham review

cwickham (commented):
I pushed some changes to incorporate @cderv's points, and then based on a more comprehensive review of the page.

cwickham:
@cderv this is looking good, do you want to take one final look and merge?

PR #1678, opened 2025-06-01
fix: use code blocks for copy button

mcanouil:
This pull request enhances the `docs/troubleshooting/index.qmd` file by adding platform-specific instructions for setting the `QUARTO_LOG_LEVEL` environment variable.

### DIFF
--- a/docs/troubleshooting/index.qmd
+++ b/docs/troubleshooting/index.qmd
@@ -85,6 +85,26 @@ export QUARTO_PRINT_STACK=true
 
 Quarto will print more information about its internal state if you set `QUARTO_LOG_LEVEL=DEBUG` in your environment.
 
+::: panel-tabset
+
+## Windows
+
+On PowerShell:
+
+```powershell
+$env:QUARTO_LOG_LEVEL = "DEBUG"
+```
+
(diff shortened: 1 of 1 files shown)


docs/troubleshooting/index.qmd
```diff
@@ -85,6 +85,26 @@ export QUARTO_PRINT_STACK=true
...
 Quarto will print more information about its internal state if you set `QUARTO_LOG_LEVEL=DEBUG` in your environment.
 
+::: panel-tabset
+
+## Windows
+
+On PowerShell:
+
+```powershell
+$env:QUARTO_LOG_LEVEL = "DEBUG"
```
cderv, line 95: I would remove the space too like on linux 
```suggestion
$env:QUARTO_LOG_LEVEL="DEBUG"
```

cderv:
> I did not get what you meant here.

Sorry then - I did not get the direct link between `fix: use code blocks for copy button` and the PR content which is not fixing something about copy button missing but adding tabsets to get more precise about how to set environment on Linux and Windows. 

I guess I am used to see you use very specific commit message with prefix - so it suprised me 😅

PR #1685, opened 2025-06-07
Remove duplicate 'Unicode Characters' section

yocoldle:


cderv:
Thanks this is indeed an oversight of 
- https://github.com/quarto-dev/quarto-web/pull/1161
that was supposed to move the section, while expanding it.

PR #1688, opened 2025-06-09
docs(markdown-basics.qmd): use proper list indentation (stricter syntax)

mcanouil:
Ensure proper indentation for lists by correcting the first character alignment.
Also add an example with a code block inside a list.


mcanouil:
don't merge as something is off with the `A. ` item.

cderv:
@mcanouil looks good now right ?

mcanouil:
Yes, the list has the correct structure now.

PR #1696, opened 2025-06-17
Switch to `engine: julia` in prominent examples

jkrumbiegel:
The docs already recommend the `julia` engine over `jupyter` in the julia section, so it makes sense to adjust these copy-paste-able examples to that engine as well so users are put on the right track from the start.


cderv:
Rhanks

PR #1717, opened 2025-07-21
Add a FAQ point regarding `!expr` and `!r`

cderv:
for now only `!expr` is supported

To clarify 
- https://github.com/quarto-dev/quarto-cli/discussions/13071#discussioncomment-13825655


fkohrt:
> `!expr` is handled by the **yaml** R package, while `!r` is handled by **knitr** via `xfun::yaml_load()`.

I think the second part is not correct: `!r` is handled by `knitr::knit_params()` (and `!expr` is handled by `xfun::yaml_load()` and `yaml::yaml.load()`, subsequently).

PR #1723, opened 2025-07-24
Update callouts.qmd

serialc:
Added the 'caution' callout example for completeness and explained the heading behaviour

### DIFF
--- a/docs/authoring/callouts.qmd
+++ b/docs/authoring/callouts.qmd
@@ -31,6 +31,12 @@ Callouts provide a simple way to attract attention, for example, to this warning
 Danger, callouts will really improve your writing.
 :::
 
+::: callout-caution
+## Headings are not necessary
+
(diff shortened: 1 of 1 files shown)


cderv:
Callout caution is below : 
https://github.com/quarto-dev/quarto-web/blob/f3cd2b83bb7380166268f477cf2c4ee4814d798c/docs/authoring/callouts.qmd#L40-L44

It is just collapse by default to show this example
image

cderv:
We know get one more callout than the number of type
image

> Added the 'caution' callout example for completeness

So I am not sure this was missing. 

> and explained the heading behaviour

However, we could indeed show example of that - the two first are example, but they don't explain it. 

So we could review it and adapt, but having two cautions is maybe too much

serialc:
I removed redundant content, clarified the headings issue, and added emphasis for the "Caution" callout.

PR #1731, opened 2025-07-29
Add explanation on conditional code execution

Joao-O-Santos:
The quarto documentation clearly explains how to add sections to a quarto document that are only displayed in a given input format. Yet, it is my understanding that the code in code chunks that are in those sections is always evaluated.

There are times where one would wish to have the code in those chunks only be evaluated/executed when the document is being rendered to that specific format. This happened to me when writing a document where I wanted to print a `DT::datatable()` when rendering to HTML and a `knitr::kable()` when rendering to any other format. Having the `DT::datatable()` chunk execute when rendering to a format other than HTML caused errors.

I found previous issues, discussions and comments addressing this, so I think I'm not the only one experiencing this problem:
- [3260](https://github.com/quarto-dev/quarto-cli/discussions/3260)
- [11724](https://github.com/quarto-dev/quarto-cli/issues/11724) with this [comment](https://github.com/quarto-dev/quarto-cli/issues/11724#issuecomment-2557635192) explaining the solution.

I reckon this could be clearly and concisely explained in the documentation.

This pull request is my attempt to do just that.
I'm open to refactoring it and revising per your feedback.

Best,
J

P.S: congratulations and a giant thank you for building quarto!
I confess I was a bit skeptical at first about moving to quarto from RMarkdown, but when I started reading the quarto documentation I experienced just how much you streamlined the project rendering experience, I was able to streamline quite complicated Makefiles, and have everything more accessible to less experience collaborators.
I particularly appreciate how your documentation is editor agnostic, and does not push RStudio, and love the fact that you thought of us `neovim` users :)


docs/authoring/conditional.qmd
```diff
@@ -25,6 +25,16 @@ You can also set conditions on non-executable code blocks:
 ```
 ````
 
+When using the `knitr` engine, if you want to have a code chunk being excecuted/evaluated only when the output is being rendered to HTML or LaTeX you can use the `is_html_output()` and `is_latex_output()` respectively (see [this reply to quarto-cli issue 11724](https://github.com/quarto-dev/quarto-cli/issues/11724#issuecomment-2557635192)).
```
cderv, line 28:
```suggestion
When using the `knitr` engine, if you want to have a code chunk being executed/evaluated only when the output is being rendered to HTML or LaTeX you can use the `is_html_output()` and `is_latex_output()` respectively (see [this reply to quarto-cli issue 11724](https://github.com/quarto-dev/quarto-cli/issues/11724#issuecomment-2557635192)).
```

cderv (commented):
Thanks for contributing a PR. 

This is indeed an interesting example. The Quarto documentation aims to be as general as possible, so I worry this addition is too specific for now. It also show a use of `!expr` which is also knitr specific. 

I wonder if we should have a specific small blog post on this trick, or an online example we could link to from the main doc. 
Or a collapsible callout ? 

Once 
- https://github.com/quarto-dev/quarto-cli/issues/2216
is solved, the output format will be accessible through environment variable and accessible to all engine. 

When this is available, then this would have a good place in the main code I believe.

cderv:
Thanks for the new version. 

I believe we will document this maybe in a blog post, or an example as in 1.8 we'll have a way to do that in Python too. 

- https://github.com/quarto-dev/quarto-cli/issues/2216

So I think it will be best placed elsewhere and maybe linked... 

I'll let @cwickham make the final decision if it fits

PR #1738, opened 2025-08-02
Fix typos in neovim.qmd

lbm364dl:


lbm364dl:
How do you find the typos? I don't think it's just you searching random words, is it?

lbm364dl:
I added the other fixes you suggested. For the changelog ones I checked that the messages weren't part of the linked PR, so it should be fine.

PR #1752, opened 2025-08-20
fix: `margin-header` expects markdown

mcanouil:
Update the documentation to reflect that the `margin-header` should reference a markdown file/text instead of an HTML file.

### DIFF
--- a/docs/websites/website-blog.qmd
+++ b/docs/websites/website-blog.qmd
@@ -418,16 +418,22 @@ Here are the steps required to add a subscription widget:
     The above widget HTML will not work in your blog as the subscription HTML needs to contain value specific to your blog. See your email service provider's instructions for generating a subscription widget.
     :::
 
-2.  Create the file `subscribe.html` in the root of your project.
+2.  Create the file `subscribe.md` in the root of your project.
 
-3.  Paste the HTML for your subscription widget into `subscribe.html` and save the file.
+3.  Paste the HTML for your subscription widget into `subscribe.md` as a raw HTML block and save the file.
 
-4.  Add `subscribe.html` to the `margin-header` in your `_quarto.yml` file:
+    ````{.markdown filename="subscribe.md"}
+    ```{=html}
+    Your HTML code goes here.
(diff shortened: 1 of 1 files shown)


cderv (commented):
Did you notice any issues with the feature or just documentation problem to you ? 

I believe this was documenting using .html because it made more sense as the services to extract the subscription code would produce a .html file, and the content is supposed to be HTML only, with no markdown. 

You're right that the documentation does not mention the raw html part, 
but this is because this is handled internally in the code. 

See https://github.com/quarto-dev/quarto-cli/blob/2a4799429cdfca4a055aa465bc56790d7ff0483f/src/project/types/website/website-navigation-md.ts#L716-L731

````ts
function expandMarkdownFilePath(source: string, path: string): string {
  const absPath = isAbsolute(path) ? path : join(dirname(source), path);
  if (safeExistsSync(absPath)) {
    const fileContents = Deno.readTextFileSync(absPath);

    // If we are reading raw HTML, provide raw block indicator
    const ext = extname(absPath);
    if (ext === ".html") {
      return "```{=html}\n" + fileContents + "\n```";
    } else {
      return fileContents;
    }
  } else {
    return path;
  }
}
````

This logic applies for all header / footer (margin, body, sidebar)

So passing a .html file directly to `margin-header` is supposed to work ok and content will be wrapped in raw html block. 

So I don't think the documentation should be adapted here.

mcanouil:
> It does not have the same code path as the listing, and there is no EJS templating here, so there will be a difference in documentation. Is this backed by a Github discussion or issue that shows the confusion?

No, but my point is that using raw HTML in markdown without raw block leads to issues (the worst being in EJS files such as listings).
The documentation states everywhere that `margin-*` is markdown as text or file. Only this particular example uses HTML without any explanation about why here and only here an HTML file is allowed eventhough the documentation says "markdown". This implies that you could try HTML in other places and that it should work perfectly, but that's not true.

I don't mind going the other way around and state that these `margin-*` support raw HTML file as well.

PR #1792, opened 2025-10-06
Remove `posit-cloud` as a publishing venue

cwickham:
* Leave `docs/publishing/posit-cloud.qmd` with no content except a callout (for people who follow a link)

* Remove mentions from navigation, and tables of publishing venues.


cderv (changes requested):
We may have a problem or conflict in ID sidebar or something because I am suprised that this page
- https://deploy-preview-1792.quarto.org/docs/publishing/posit-cloud.html

has the Quarto 1.4 Highlight sidebars... 
image

Or it may just be because the page is removed from the `_quarto.yml`
````yaml
            - section: "Publishing"
              contents:
````

and so another sidebar is automatically assigned to it. (I believed sidebar are attributed based on content section)

So we may need to fix this before merging

PR #1810, opened 2025-10-11
fix: update issue URL to use documentation template

mcanouil:
Update the issue URL in the configuration to direct users to the documentation issue template, enhancing guidance for reporting issues.

There is no reason to ask users to choose which template to use (including "blank issue") when they click on "report an issue" from https://quarto.org.
Let's guide them directly to the right one.


cderv (approved):
Thank you. Sorry for the delay !

PR #1822, opened 2025-10-31
Add url-card shortcode

multimeric:


multimeric:
Well I prefer QuartoCards as a repo name, in case I create more shortcodes. But I note there are a few repos like https://github.com/gadenbuie/quarto-partials and https://github.com/mcanouil/quarto-modal where the `quarto-` part is stripped off in the extensions page. This is what I was aiming for, because having Quarto in this list is a bit redundant. However, if your current policy is to not allow this, I might just rename the repo to `quarto-cards` to align with the style at least.

cderv:
> I create more shortcodes.

I meant that if you do add more shortcode to it, we wouldn't create a new entry I guess. Or at least for now we do list repo once. 

Idea of this listing is to list the existing extensions and not each feature they have. 

For the example you listed, the repo name is different but extension name is related
- `title: Modal` in `_extensions/modal/_extension.yml` for https://github.com/mcanouil/quarto-modal ([source](https://github.com/mcanouil/quarto-modal/blob/e43f072d2921ee1c2db2f8f94a474332b0ab9d2b/_extensions/modal/_extension.yml#L1)) 
- `title: partials` in `_extensions/partials/_extension.yml` for https://github.com/gadenbuie/quarto-partials ([source](https://github.com/gadenbuie/quarto-partials/blob/73366b1dee25b9ee83af583bb846dfe12c5aed57/_extensions/partials/_extension.yml#L1))

In your case we have `title: QuartoCards` in `_extensions/QuartoCards/_extension.yml` ([source](https://github.com/WEHI-ResearchComputing/QuartoCards/blob/c36370858d11aa5db6d4442476966b726506ede3/_extensions/QuartoCards/_extension.yml#L1)). So extensions is really `QuartoCards` when you install it. 

Hence my initial questions. For us, this would be one extension in the listed that could offer multiple shortcode. Does this clarify my initial comment ? 

But I can see why listing shortcode name instead of extension could be useful for search... 🤔

cderv (changes requested):
For our current way to list, we prefer to list by extension name, so it would be QuartoCards. 

Can I change it in the listing ? Do you want to rename ? 

We may evolve in the future to have listing by shortcode of filter name, but for now this is by extension name. 

Also you did not have a licence in your repo. Would be best to add one. 

Thank you!

multimeric:
Okay, I've updated it to the extension name. Is it an issue that I'm PascalCase? Because I can rename the extension and repo to snake_case if that is preferable

cderv:
I don't think it is . CLearly not the convention, but it will still install

docs/extensions/listings/shortcodes-and-filters.yml
```diff
@@ -882,9 +882,8 @@
 
 - name: QuartoCards
   path: https://github.com/WEHI-ResearchComputing/QuartoCards
-  author: '[Michael Milton](https://github.com/multimeric)'
-  description: >
-    Generate cards for displaying URLs in your Quarto presentations.
+  author: '[multimeric](https://github.com/multimeric)'
```
cderv, line 885: Just curious: You prefer to hide you name in the listing ?

multimeric: That's right

PR #1892, opened 2026-02-20
Update style guide 

cwickham:


cderv:
I think we should consider writing a 🤖 rule that points to that, or really migrate this to a 🤖 rule directly.

PR #2003, opened 2026-04-29
Add Reference sidebar landing pages to fix breadcrumb navigation

cderv:
When clicking the `Formats` breadcrumb on `/docs/reference/formats/<page>.html`, the link led to `formats/html.html` (the first child) instead of a section landing. The `Reference > More` section had a similar issue and was not a coherent grouping for its contents.

Adds explicit landing pages for three Reference sidebar sections so the breadcrumb renders as a clickable link to a real page:

- `Formats` → new `docs/reference/formats/index.qmd`
- `Projects` → new `docs/reference/projects/index.qmd`
- `More` → renamed to `Metadata` and pointed at new `docs/reference/metadata/index.qmd`

The `More` → `Metadata` rename matches the grouping already used in `docs/reference/reference.yml` (the master listing the website renders) — Dates, Globs, Citations, Cross-References, and Brand are now grouped under `Metadata` in the sidebar too. The five items themselves are unchanged.

Each landing mirrors the existing `docs/reference/cells/` pattern: small `index.qmd` shell + a per-subdir `reference.yml` subset rendered by the existing `ejs/links.ejs` template. No new extension, shortcode, or filter. The master `docs/reference/reference.yml` is untouched.

Reported in quarto-dev/quarto-cli#14402. The underlying CLI fallback behavior (sidebar sections without `href:` use `contents[0].href`) is tracked separately at quarto-dev/quarto-cli#14454 and is not addressed here. Inner Guide sub-sections (HTML/PDF/MS Word/Typst/Markdown under Documents) and `Reference > Quarto CLI` continue to inherit the same fallback and are deferred.

A follow-up PR will fix the `Documents` section in the Guide sidebar.

### DIFF
--- a/_quarto.yml
+++ b/_quarto.yml
@@ -493,6 +493,7 @@ website:
           href: docs/reference/index.qmd
           contents:
             - section: "Formats"
+              href: docs/reference/formats/index.qmd
               contents:
                 - text: "HTML"
                   href: docs/reference/formats/html.qmd
--- a/_quarto.yml
+++ b/_quarto.yml
@@ -597,7 +597,8 @@ website:
                   href: docs/reference/projects/books.qmd
                 - text: "Manuscripts"
                   href: docs/reference/projects/manuscripts.qmd
-            - section: "More"
+            - section: "Metadata"
+              href: docs/reference/metadata/index.qmd
               contents:
                 - text: "Dates"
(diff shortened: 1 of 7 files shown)


cwickham (approved):
This looks great to me.

PR #2015, opened 2026-05-06
add journal template for Cambridge University Press

ihrke:
https://github.com/ihrke/quarto-cambridge


cderv:
@ihrke when you create a PR, best is to do it using a branch and not from main directory. Just an advice. Thanks for the submitting !

PR #2038, opened 2026-05-20
docs: document quarto.* template variable namespace

cderv:
Document the new `quarto.*` Pandoc template-variable namespace, which exposes Quarto's resolved localized strings (the merged result of `_language.yml` + `lang:` locale + `language:` overrides) to custom templates and template partials as `$quarto.language.<key>$`.

Adds:

- `docs/authoring/language.qmd` — "Localized Strings in Templates" section explaining the namespace, with a custom-template example and an internal-namespace callout
- `docs/journals/templates.qmd` — short pointer section linking back to the canonical docs from the template-authoring page
- `docs/prerelease/1.10/_highlights.qmd` — v1.10 highlight bullet

All new content is gated by `{{< prerelease-callout 1.10 >}}`, so the sections stay hidden in published docs until v1.10 ships.

Companion to quarto-dev/quarto-cli#14530


cderv:
This will need to wait for next prerelease builds to merge

PR #2043, opened 2026-05-26
docs: document conditional execution via QUARTO_EXECUTE_INFO

### LINKED ISSUE #13514: Document engine-agnostic approach to conditional execution
With `QUARTO_EXECUTE_INFO`, we now have an approach to conditional execution that works across engines.

For example, for code that should be only executed for HTML formats:

R:
````markdown
```{r}
execute_info <- jsonlite::read_json(Sys.getenv("QUARTO_EXECUTE_INFO"))
is_html <- execute_info[["format"]][["identifier"]][["base-format"]] == "html"

if (is_html) {
  "HTML"
} 
```
````

Python:
````markdown
```{python}
import json
import os

with open(os.getenv("QUARTO_EXECUTE_INFO")) as f:
    execute_info = json.load(f)

is_html = execute_info["format"]["identifier"]["base-format"] == "html"

if is_html:
    print("HTML")
```
````

Users often want to combine this with [conditional content](https://quarto.org/docs/authoring/conditional.html).

This could be documented:
 
* as a section of [`/docs/computations/execution-options.qmd`](https://quarto.org/docs/computations/execution-options.html)
* its own "Conditional Execution" page under Guide > Computations 
* as an example wherever the docs for `QUARTO_EXECUTE_INFO` end up
* as a blog post on "Conditional Execution"

Regardless of where the examples end up, it should be mentioned and linked to from:

* Guide > Computations > Using Python
* Guide > Computations > Using R
* Guide > Computations > Using Julia
* Guide > Authoring > Advanced > Conditional Content
* https://quarto.org/docs/authoring/cross-references-divs.html#conditional-content
* Wherever the docs for `QUARTO_EXECUTE_INFO` end up

Some relevant issues and discussions this would resolve:
 
* https://github.com/quarto-dev/quarto-cli/discussions/3260
* https://github.com/quarto-dev/quarto-cli/issues/11724
* https://github.com/quarto-dev/quarto-web/pull/1731

### PR DESCRIPTION
**Pull request by cderv**
Cross-link the existing `QUARTO_EXECUTE_INFO` reference page (`docs/advanced/quarto-execute-info.qmd`) from the six pages users actually browse when looking for conditional-execution behavior. The reference page itself is not modified.

Per the issue thread on quarto-dev/quarto-cli#13514, users want both:

- a single canonical place documenting the env var (`docs/advanced/quarto-execute-info.qmd`, already exists)
- concise, language-specific examples on the pages users land on (R, Python, Julia)

This branch addresses the second part — the first already exists.

## Changes

- `docs/computations/execution-options.qmd` — new `## Conditional Execution` section explaining the distinction from format-based conditional content.
- `docs/computations/r.qmd`, `python.qmd`, `julia.qmd` — `## Conditional Execution` section per language, with a short verbatim snippet showing how to read `QUARTO_EXECUTE_INFO`. The R section also points at `knitr::is_html_output()`, `knitr::is_latex_output()`, and `knitr::pandoc_to()` for users who only need the common cases.
- `docs/authoring/conditional.qmd` — explicit clarifier that `.content-visible` / `.content-hidden` hide output but do not skip execution (per cwickham's ask in the thread).
- `docs/authoring/cross-references-divs.qmd` — same execution clarifier, plus a note pointing at project profiles as another axis for content variation beyond `when-format`.
- `docs/authoring/_visibility-vs-execution.md` — shared partial used by both authoring pages so the execution-vs-visibility note has a single source of truth.

The added language snippets contain no executable code (wrapped in verbatim markdown fences with `{{lang}}` escaped cell markers), so no `_freeze/` updates are introduced.

## Test Plan

- [ ] Render each modified page; confirm new sections appear in the page TOC and link targets resolve.
- [ ] Confirm `cross-references-divs.qmd` Conditional Content section now ends with the two notes (execution + project profiles) plus the existing example.
- [ ] Confirm `_visibility-vs-execution.md` partial renders identically in both consumer pages.
- [ ] No `_freeze/` regeneration triggered.

Closes quarto-dev/quarto-cli#13514

### CODE REVIEW

File: docs/computations/julia.qmd [hunk 1]
```diff
@@ -543,6 +543,25 @@ Please direct questions and requests regarding this functionality to the
 [QuartoNotebookRunner](https://github.com/PumasAI/QuartoNotebookRunner.jl)
 repository.
 
+## Conditional Execution
```
cwickham, line 546: This ends up nested under "Using the `julia` engine" but it applies to both engines. Either repeat it, or move it up a level.

docs/computations/julia.qmd
```diff
@@ -543,6 +543,25 @@ Please direct questions and requests regarding this functionality to the
...
+
+To execute a cell only for certain output formats, read the [`QUARTO_EXECUTE_INFO`](/docs/advanced/quarto-execute-info.qmd) environment variable inside the cell:
```
cwickham: This applies to all the `docs/computations/<language>.qmd` files: I think we can add a bit more explanation of what the code does.

E.g.
```suggestion
To execute a cell only for certain output formats, read the [`QUARTO_EXECUTE_INFO`](/docs/advanced/quarto-execute-info.qmd) environment variable inside the cell.
For example, you could execute code only if the target format is based on HTML (e.g. for `html`, `revealjs`, and `dashboard` formats):
```

docs/authoring/cross-references-divs.qmd
```diff
@@ -438,3 +438,7 @@ Scatterplot
 
 @fig-scatterplot
 ````
+
+{{< include _visibility-vs-execution.md >}}
+
+When the variation you need is broader than output format — for example, the same document rendered as a draft vs. a final, or for an internal vs. an external audience — [project profiles](/docs/projects/profiles.qmd) can set alternate metadata per profile that `.content-visible when-meta=...` reads. This gives you another axis for conditional content beyond `when-format`.
```
cwickham:
```suggestion
When you want visibility to depend on something beyond output format — for example, the same document rendered as a draft vs. a final version, or for an internal vs. an external audience — combine `.content-visible`/`.content-hidden` with the `when-meta` attribute.
You can then control visibility with values you set in your metadata, either in your document header, or via [project profiles](/docs/projects/profiles.qmd).
```

docs/computations/execution-options.qmd
```diff
@@ -498,3 +498,11 @@ echo "foo"
 ````
 
 Note that the Knitr engine also supports ```` ```{python} ```` cells, enabling the combination of R, Python, and Bash in the same document
+
```
cwickham, line 501: I'm not seeing this edit...needs `_freeze` I think.

cwickham (changes requested):
Edits look great! 

Is this comment in the PR from Claude?:

> The added language snippets contain no executable code (wrapped in verbatim markdown fences with {{lang}} escaped cell markers), so no _freeze/ updates are introduced.

The reasoning seems wrong, i.e. even if includes don't introduce executable code, we still need to update `_freeze/` if the original page has executable code. Might be worth a rule/instruction somewhere for Claude...

In this case `docs/computations/execution-options.qmd` has executable code and does need a re-render, commit to `_freeze/`.

cderv:
Yes, that comment was from Claude — and the reasoning was wrong. I should have catch it. Thank you ! 

To avoid this in the future, I added two enforcement layers to the branch:

- `.githooks/pre-commit` — blocks commits where a staged `.qmd` with a `_freeze/` entry has a stale hash. Activate with `git config core.hooksPath .githooks`.
- `.claude/settings.json` — a Claude Code PreToolUse hook that intercepts `git commit` and `git push` in Claude sessions and runs the same check.

Both delegate to:

https://github.com/quarto-dev/quarto-web/blob/09e75b8fc1b221d9b53f503a7dca59b67b5831bc/.githooks/check-freeze.sh

The rule in `.claude/rules/quarto-web-workflow.md` was also updated — `quarto render <file.qmd>` is now the prescribed fix when the hook fires.

Hopefully this works on Mac too, and will help us and claude to know when to update. 

I wish to have a 'quarto inspect <doc>' output that tell us if the document is frozen. We could then always re-render.

PR #2050, opened 2026-06-03
Fix broken redirect code

nrennie:
At the bottom of https://quarto.org/docs/get-started/ there's link to https://quarto.org/docs/get-started/hello/ which doesn't correctly redirect and shows a mostly empty page. I think it's just a missing closing comment symbol that means the redirect isn't actually executing.


cderv (approved):
Thank you @nrennie ! 

And thank you @cwickham for the correct trigger 🤦 

Also I have push a fix for the wrong preview file list.

PR #2086, opened 2026-06-27
fix(filters): fix include-dark with meta-resolved image paths

### LINKED ISSUE #14583: Image missing https://quarto.org/docs/computations/python.html#positron

Report an issue on quarto.org

### Description

Discovered while reviewing another PR. Opening to remind me to fix.

mcanouil:
The Positron screenshot was broken on the Python, R, and Julia computations pages and images showed as missing.

The root cause is in how Quarto encodes shortcodes before handing the document to Pandoc. See quarto-dev/quarto-cli#14583 for details.

The filter now checks for the double-extension case (same extension appearing twice) and strips the outer spurious one before building the dark and light image paths. This makes the filter robust for any `{{< meta >}}` + `.include-dark` combination until the root cause is fixed upstream in the qmd reader. This is more a workaround than an actual fix, but it at least fix the issue in the documentation with minimal change.

Closes quarto-dev/quarto-cli#14583

### DIFF
--- a/filters/include-dark.lua
+++ b/filters/include-dark.lua
@@ -3,7 +3,7 @@ local function dark_path(src)
   return path[1] .. "-dark" .. path[2]
 end
 
-local function new_image(img, src, class) 
+local function new_image(img, src, class)
   local new_classes = pandoc.List.filter(img.classes, function(c)
     return c ~= "include-dark"
   end)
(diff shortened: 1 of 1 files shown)


cderv:
The root fix is landing in quarto-dev/quarto-cli#14634. Once that ships, the doubled extension will no longer reach the filter, so this workaround PR can simply be closed without merging.

cderv:
I think new prerelease fixed it. 
image

so I'll close. Thanks for the PR!

PR #2088, opened 2026-06-30
docs(authoring): document backslash math delimiters

### LINKED ISSUE #12263: Document support for ` \(` and `\[` math delimiters

Report an issue on quarto.org

### Description

Originally suggested by @JamesSmith20, moved from https://github.com/quarto-dev/quarto-cli/issues/11761 :

#### 2. **Support for `\(` and `\[` Math Delimiters**
Currently, the [Quarto documentation on equations](https://quarto.org/docs/visual-editor/technical.html) does not mention support for LaTeX-style math delimiters `\(`...`\)` (inline) and `\[...\]` (block). It states that only `$...$` and `$$...$$` are supported for inline and block math. However, using Pandoc’s `markdown+tex_math_single_backslash` extension, Quarto can support these delimiters seamlessly across Word, HTML, and PDF formats.

Here is an example:

```markdown
---
title: "Quarto Playground"
from: markdown+tex_math_single_backslash
format:
  docx
---

Inline math example: \( E = mc^2 \)

Inline math example: $E = mc^2$

Block math example:

\[ a^2 + b^2 = c^2 \]

$$
a^2 + b^2 = c^2
$$
```

This works well to render equations using the preferred LaTeX-style delimiters, which are consistent with LaTeX standards and widely used by large language models and other AI tools. Including this information in the official documentation would help users integrate equations more easily without manual adjustments.

### Suggested Improvements:

2. **Document Support for `\(` and `\[` Delimiters**:
   - Update the math documentation to mention the use of `markdown+tex_math_single_backslash` for LaTeX-style math delimiters.
   - Include examples to show how `\(`...`\)` and `\[...\]` can be used for inline and block math.

### Benefits:
- Makes Quarto’s capabilities more transparent to users.
- Enhances usability for academic and technical writing by providing official guidance on:
  - LaTeX-style math delimiters for consistency.
---

### Related Discussions:
This request builds upon [Discussion #11753](https://github.com/quarto-dev/quarto-cli/discussions/11753#discussioncomment-11702276) on Quarto’s GitHub platform.

Thank you for considering this enhancement to the documentation!

mcanouil:
Add a note to the Equations section of the Markdown Basics page documenting that `\(...\)` and `\[...\]` math delimiters are supported via Pandoc's `tex_math_single_backslash` extension.

Closes quarto-dev/quarto-cli#12263

### DIFF
--- a/docs/authoring/markdown-basics.qmd
+++ b/docs/authoring/markdown-basics.qmd
@@ -387,7 +387,27 @@ $$
 :::
 ```
 
-For HTML math processed using [MathJax](https://docs.mathjax.org/) (the default) you can use the `\def`, `\newcommand`, `\renewcommand`, `\newenvironment`, `\renewenvironment`, and `\let` commands to create your own macros and environments.  
+For HTML math processed using [MathJax](https://docs.mathjax.org/) (the default) you can use the `\def`, `\newcommand`, `\renewcommand`, `\newenvironment`, `\renewenvironment`, and `\let` commands to create your own macros and environments.
+
+You can also use LaTeX-style `\(...\)` delimiters for inline math and `\[...\]` for display math by enabling the `tex_math_single_backslash` Pandoc extension via the `from` option:
+
+``` yaml
+from: markdown+tex_math_single_backslash
+```
+
++-------------------------------+-------------------------+
+| Markdown Syntax               | Output                  |
++===============================+=========================+
+| ``` markdown                  |                         |
+| inline math: \(E = mc^{2}\)  | inline math: $E=mc^{2}$ |
+| ```                           |                         |
++-------------------------------+-------------------------+
+| ``` markdown                  |                         |
+| display math:                 | display math:           |
+|                               |                         |
+| \[E = mc^{2}\]                | $$E = mc^{2}$$          |
+| ```                           |                         |
++-------------------------------+-------------------------+
 
 ## Diagrams
 


cderv:
This `from` option is tricky because it is kind of hidden (probably a bit on purpose as not sure every option works well - thinking this because of our custom lua reader). 

I wonder if somehow this means that we should expose this type of options differently... 🤔 

To avoid users to set `from` but use another option. Like `math-delimiter` ? 

Other related thinking: 

This syntax Pandoc extension syntax can also be confusing because (1) this is pandoc feature, (2) we don't explain it anywhere I think - except through our issue replies when needed.  So just doing it for this specific `+tex_math_single_backslash` is maybe too out of the box. I would try see if we can add context about this is being pandoc  (https://pandoc.org/MANUAL.html#math-input) and think on how to explain it. 

@cwickham did you have something in mind when opening https://github.com/quarto-dev/quarto-cli/issues/12263 ? Any thought on this ?

mcanouil:
That's true it's kind of hidden, but in plain sight as it's suggested by the YAML schema, so you can see it in auto-complete, etc.

And it's already mentioned in the documentation for the visual editor section for emoji support: https://quarto.org/docs/visual-editor/content.html#emojis

We could add a from callout/section in markdown-basics.qmd to explain the `from` reader option. Maybe something about variants as well, as it's a recurring thing we talk about in issues/discussion, it might deserve more space in the documentation, but this is a separate task/PR I believe.

cderv:
yes visual editor doc has been 'copied' into quarto.org without much adaptation from https://rstudio.github.io/visual-markdown-editing/

> Maybe something about variants as well, as it's a recurring thing we talk about in issues/discussion

yes i know - it 'kinda' works but we know it is not full proof. Being undocumenting help reduce the surface of possible problem :) 

> it might deserve more space in the documentation, but this is a separate task/PR I believe.

Agreed this is another PR. I wondered about the order. We could document reader and writer logic, and `from` + `to` and `variant` and be able to link from other part like this one. 

🤷 I don't know really - maybe this is fine to put this `+tex_math_single_backslash` out there without doing more, but I feel this should be simpler if user's needs that. We try to hide most Pandoc logic, expecially when complex, so we could do that here with a new math config.

cwickham:
In regards to https://github.com/quarto-dev/quarto-cli/issues/12263, I was just splitting an issue in two. I wasn't sure how much we want to expose Pandoc extensions. Will q2 be a bit more opinionated on this?

cderv:
> Will q2 be a bit more opinionated on this?

Oh I think it will. We don't use Pandoc binary so I don't know if we are really supporting all pandoc feature like this. non default quarto one. 

So another way to maybe not document this that much if we don't support all.

I need to verify, but possible this Math syntax won't be supported. @cscheid did you already decide on how q2 will behave related to pandoc format extensions ?

cscheid:
> I need to verify, but possible this Math syntax won't be supported. @cscheid did you already decide on how q2 will behave related to pandoc format extensions ?

Yes. We support only what our dialect does and won't have extensions at all (yes, this will upset some users). If there are sufficiently large gaps in user expectations and what we can parse, we will change the qmd parser and _everyone_ gets it. We can't afford to compile a large number of tree-sitter-qmd variants and switch between them arbitrarily, and we need a high-performance parser for the LSP (source diagnostics) and incremental-writer applications (quarto-hub).

cscheid:
I actually would propose to _not_ document this here, because it's likely going away for q2...

cderv:
> I actually would propose to not document this here, because it's likely going away for q2...

That is what I thought. I agree with it. 

Users who want this in quarto 1 can find info searching issues today. Not great, but our replies to issues are still documentation in a way.

PR #2100, opened 2026-07-12
Update shortcodes-and-filters.yml to add qcode extension

skyfroger:
qCode is an extension for Quarto that adds an interactive Python IDE to your page with the ability to run code directly in the browser, as well as search documentation for Python functions and methods.


cderv (approved):
Thank you

PR #2120, opened 2026-07-24
Add dark-mode images for figures and markdown-basics docs

cderv:
Figures and Markdown Basics ship light-only elephant screenshots, which look jarring against a dark background when readers toggle the site's dark theme.

Documents can now show a dark-mode variant for each image, using the existing `.light-content`/`.dark-content` convention already used elsewhere on the site.

Supersedes #1641, which added these dark assets back in April 2025 but went stale. That PR vendored a `pandoc-ext/list-table` Lua filter to mix code blocks and images in table cells; this rewrites that markup using the list-table syntax Quarto 1.9 shipped natively, so no extension is needed.

Also wires up `layout-attrib-negative-dark.png`, an asset #1641 added but never referenced.


cderv:
Preview checked on deploy build (https://deploy-preview-2120.quarto.org) — figures, markdown-basics, tables, cross-references pages. Verified light/dark image swap via computed style, and confirmed no broken image links (all 200s). Closed #1641 as superseded.

PR #2140, opened 2026-08-12
feat(gallery): redesign gallery with intent-driven landing

mcanouil:
Replace the carousel gallery with an intent-driven landing plus six category pages. The landing has a category chooser, a section rail, and curated featured examples with animated hover previews. Each category page lists its examples with keyboard-operable filters, sort, and search, and a browse page aggregates everything with chips derived from the data.

Cards are driven by YAML under `docs/gallery/data/`, thumbnails consolidated under `thumbnails/` with capture tooling documented in READMEs, and the interactive chips meet WCAG AA. The legacy gallery is kept under `docs/gallery/legacy/` (referenced as a working example for custom listing).


mcanouil:
Note: need to drop https://github.com/noahweidig/noahweidig.github.io as it's now using Astro

PR #2144, opened 2026-08-15
Use `actions/checkout@v7` in github-pages.qmd

guslipkin:
`actions/checkout@v4` now results in the following message. `actions/checkout@v7` is the most recent major version and does not have the message.

`
Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: actions/checkout@v4 . For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
`

### DIFF
--- a/docs/publishing/github-pages.qmd
+++ b/docs/publishing/github-pages.qmd
@@ -165,7 +165,7 @@ jobs:
       contents: write
     steps:
       - name: Check out repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v7
 
       - name: Set up Quarto
         uses: quarto-dev/quarto-actions/setup@v2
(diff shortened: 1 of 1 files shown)


cscheid:
Thanks, but as a matter of safety policy we don't take external PRs for our CI infrastructure (or change suggestions to those policies). We'll do a separate review.

cscheid:
@cderv I'm 99% sure this is safe for our docs to change, but just FYI.

cscheid:
[We ourselves use `@v6`](https://github.com/quarto-dev/quarto-web/blob/main/.github/workflows/publish.yml#L30C1-L30C34); once again I'm 99% confident this is fine! But, you know.

cderv:
Fixed in https://github.com/quarto-dev/quarto-web/pull/2150

PR #2148, opened 2026-08-18
Add Oxford Transfer custom format

kv9898:
Adds the Oxford Transfer of Status Quarto PDF format to the custom-format extension listing.

The format is an unofficial adaptation of the Oxford Transfer of Status template, with attribution and protected-crest guidance in its repository.


cderv (approved):
Thanks for the submission! Everything checks out:

- Repo is public with a clear license (CC BY 4.0), and the original template's attribution (Overleaf source, University crest handling) is documented carefully in `NOTICE.md`/`LICENSE`.
- README has both a `quarto use template` and a `quarto add` command, plus a full usage example.
- Entry is correctly categorized and alphabetically placed in `custom-formats.yml`.
- Extension installs cleanly with `quarto add`.
- No scripting in the extension — just LaTeX template partials — so no security concerns.

Approving.

PR #2150, opened 2026-08-23
docs: set github-actions steps version as variables

### LINKED ISSUE #14796: GitHub Pages publishing docs still show `actions/checkout@v4`, triggering Node 20 deprecation warnings
### I have:

- [x] searched the issue tracker for similar issues
- [x] installed the latest version of [Quarto CLI](https://github.com/quarto-dev/quarto-cli/releases/latest)
- [x] formatted my issue following the [Bug Reports guide](https://quarto.org/bug-reports.html)

### Bug description

The [GitHub Pages publishing guide](https://quarto.org/docs/publishing/github-pages.html#publish-action) currently pins `actions/checkout@v4` in all three of its workflow examples (Publish Action, Jupyter+venv, Knitr+renv). Following these docs as written now surfaces GitHub's Node 20 deprecation warning, since `actions/checkout@v4` runs on the Node 20 runtime that [GitHub retired on Actions runners](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/).

Note that `quarto-cli`'s own CI moved off `actions/checkout@v4` in [#14199](https://github.com/quarto-dev/quarto-cli/pull/14199) (v4 → v6, along with actions/cache, upload-artifact, download-artifact, and setup-node), citing this same Node 20 deprecation.

**Already fixed** is the [quarto-actions/publish README](https://github.com/quarto-dev/quarto-actions/tree/main/publish#github-pages); this already carries a note that the action is compatible with actions/checkout@v6+, with @v5 as a fallback pin (referencing [quarto-actions#133](https://github.com/quarto-dev/quarto-actions/issues/133)) — but that's a repo README, not a page on quarto.org, so it's easy to miss.

**Ask:**

Update the three actions/checkout@v4 references on the GitHub Pages publishing page to @v5 or @v6 so the copy-pasteable examples don't hand new users a deprecation warning. 

**Thanks Quarto Team!!**

### Steps to reproduce

Use the copy/pasted publish workflow from Quarto's [GH pages publishing docs](https://quarto.org/docs/publishing/github-pages.html#publish-action):
```
on:
  workflow_dispatch:
  push:
    branches: main

name: Quarto Publish

jobs:
  build-deploy:
    runs-on: ubuntu-latest
    permissions:
      contents: write
    steps:
      - name: Check out repository
        uses: actions/checkout@v4

      - name: Set up Quarto
        uses: quarto-dev/quarto-actions/setup@v2

      - name: Render and Publish
        uses: quarto-dev/quarto-actions/publish@v2
        with:
          target: gh-pages
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

Use this to publish a quarto website and you get the Node20 warning.

### Actual behavior

Warning surfaced.

### Expected behavior

No warning surfaced. 

### Your environment

- Linux Mint 22.1 with Cinnamon
- Ubuntu 24.04 Noble pakcage base
- Positron IDE

### Quarto check output

```bash
$ quarto check
Quarto 1.10.18
[✓] Checking environment information...
      Quarto cache location: /home/cwiegand/.cache/quarto
[✓] Checking versions of quarto binary dependencies...
      Pandoc version 3.10.0: OK
      Dart Sass version 1.101.0: OK
      Deno version 2.7.14: OK
      Typst version 0.15.1: OK
[✓] Checking versions of quarto dependencies......OK
[✓] Checking Quarto installation......OK
      Version: 1.10.18
      Path: /opt/quarto/bin

[✓] Checking tools....................OK
      TinyTeX: (external install)
      Chrome Headless Shell: (not installed)
      VeraPDF: (not installed)

[✓] Checking LaTeX....................OK
      Using: TinyTex
      Path: /home/cwiegand/.TinyTeX/bin/x86_64-linux
      Version: 2026

[✓] Checking Chrome Headless....................OK
      Chrome:  (not detected)

[✓] Checking basic markdown render....OK

[✓] Checking R installation...........OK
      Version: 4.6.1
      Path: /usr/lib/R
      LibPaths:
        - /usr/local/lib/R/site-library
        - /usr/lib/R/site-library
        - /usr/lib/R/library
      knitr: 1.51
      rmarkdown: 2.31

[✓] Checking Knitr engine render......OK

[✓] Checking Python 3 installation....OK
      Version: 3.12.3
      Path: /usr/bin/python3
      Jupyter: (None)

      Jupyter is not available in this Python installation.
      Install with python3 -m pip install jupyter

[✓] Checking Julia installation...
```

mcanouil:
Setup variables to keep the github actions steps up to date with a single change to one file.

Covered steps: `actions/checkout`, `quarto-dev/quarto-actions/setup`, `quarto-dev/quarto-actions/publish`, `actions/setup-python`, `r-lib/actions/setup-r`, and `r-lib/actions/setup-renv`. The Python and R versions that the example workflows pin are variables too.

Each value is bumped to the current stable release: `actions/checkout@v7`, `actions/setup-python@v7`, Python `3.14`, and R `4.6.1`. The other steps stay at `v2`, which is still their latest major.

Note that there is a "tool" also using an outdated version of the checkout step `_tools/screenshots/examples/quarto-demo/.github/workflows/publish.yml`. That directory is a git subtree of quarto-dev/quarto-demo, so the fix belongs upstream.

Resolves quarto-dev/quarto-cli#14796

### DIFF
--- a/_variables.yml
+++ b/_variables.yml
@@ -2,3 +2,5 @@ rstudio:
   min_version: "v2022.07"
 github-actions:
   checkout: v6
+  quarto-setup: v2
+  quarto-publish: v2
--- a/_variables.yml
+++ b/_variables.yml
@@ -1,2 +1,4 @@
 rstudio:
   min_version: "v2022.07"
+github-actions:
+  checkout: v6
--- a/_variables.yml
+++ b/_variables.yml
@@ -1,6 +1,12 @@
 rstudio:
   min_version: "v2022.07"
 github-actions:
-  checkout: v6
+  checkout: v7
   quarto-setup: v2
   quarto-publish: v2
+  setup-python: v7
+  setup-r: v2
+  setup-renv: v2
+  versions:
+    python: "3.14"
+    r: "4.6.1"
--- a/docs/publishing/_github-action-options.md
+++ b/docs/publishing/_github-action-options.md
@@ -4,7 +4,7 @@ It's possible to have a Quarto project in a large GitHub repository, where the Q
 
 ``` yaml
 - name: Render and Publish
-  uses: quarto-dev/quarto-actions/publish@v2
+  uses: quarto-dev/quarto-actions/publish@{{< var github-actions.quarto-publish >}}
   with:
     target: {{< meta provider.id >}}
     path: subdirectory-to-use
@@ -14,7 +14,7 @@ By default, `quarto publish` will re-render your project before publishing it. H
 
 ``` yaml
 - name: Render and Publish
-  uses: quarto-dev/quarto-actions/publish@v2
+  uses: quarto-dev/quarto-actions/publish@{{< var github-actions.quarto-publish >}}
   with:
     target: {{< meta provider.id >}}
     render: false
--- a/docs/publishing/github-pages.qmd
+++ b/docs/publishing/github-pages.qmd
@@ -165,7 +165,7 @@ jobs:
       contents: write
     steps:
       - name: Check out repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@{{< var github-actions.checkout >}}
 
       - name: Set up Quarto
         uses: quarto-dev/quarto-actions/setup@v2
@@ -203,7 +203,7 @@ jobs:
       contents: write
     steps:
       - name: Check out repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@{{< var github-actions.checkout >}}
 
       - name: Set up Quarto
         uses: quarto-dev/quarto-actions/setup@v2
@@ -243,7 +243,7 @@ jobs:
       contents: write
     steps:
       - name: Check out repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@{{< var github-actions.checkout >}}
 
       - name: Set up Quarto
         uses: quarto-dev/quarto-actions/setup@v2
--- a/docs/publishing/github-pages.qmd
+++ b/docs/publishing/github-pages.qmd
@@ -209,9 +209,9 @@ jobs:
         uses: quarto-dev/quarto-actions/setup@{{< var github-actions.quarto-setup >}}
 
       - name: Install Python and Dependencies
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@{{< var github-actions.setup-python >}}
         with:
-          python-version: '3.10'
+          python-version: '{{< var github-actions.versions.python >}}'
           cache: 'pip'
       - run: pip install jupyter
       - run: pip install -r requirements.txt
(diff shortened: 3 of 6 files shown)


cderv (commented):
Having the versions in one place is much better than chasing them across the four pages, thanks for doing this.

Two things about the values though.

`actions/setup-python@v5` stays hardcoded in the same four pages, and v5 is still `using: 'node20'`, so those workflows will keep printing the deprecation warning that started this in #2144. Latest is v7.0.0, which is node24.

https://github.com/quarto-dev/quarto-web/blob/db905f61ca3b0b1bce3158e5353d9f22fff58a08/docs/publishing/github-pages.qmd#L212

Same line in `docs/publishing/netlify.qmd#L166`, `docs/publishing/quarto-pub.qmd#L187` and `docs/publishing/rstudio-connect.qmd#L301`.

We could add a `setup-python` variable next to the others and cover those too. It would be a shame to centralise the checkout step and leave this one to drift again, since it is the one still warning.

I left a second comment on the checkout value itself.

_variables.yml
```diff
@@ -1,2 +1,6 @@
 rstudio:
   min_version: "v2022.07"
+github-actions:
+  checkout: v6
```
cderv: Any reason for v6 rather than v7 here ? v6 is already node24 so it does fix the warning, but v7.0.1 is the current major, and having the docs one major behind means we come back to this soon. Did you check about this maybe ?

mcanouil: I missed this update from last month and thought v6 was the latest. (v7 was released the next day of 6.1)

mcanouil:
I generalised the approach and added the R/Python version as well.

cderv:
Also other thoughts after first review... I wonder if variables is the right tool here compared to `_metadata.yml` in publishing folder with some custom `meta` as we do in some other places like

https://github.com/quarto-dev/quarto-web/blob/db905f61ca3b0b1bce3158e5353d9f22fff58a08/docs/tools/vscode/_metadata.yml#L1-L4

this way it is scoped to some .qmd only 🤔 

I guess there is no good solution but we have two ways to solve this. So I figured I shared. 
What is you experience on those ?

mcanouil:
I personally use metadata only when 100% sure everything lives under a directory.
I try to avoid metadata because of the current behaviour in Quarto CLI. Depending on where you set them, you get deep copy/merge or not. This makes them harder to use across _quarto.yml, directory and file frontmatter, which is not at all a problem with variables.
Variables also have the advantage to be mandatory located at the root of a project, thus easy to locate for maintenance, no merge consideration, etc. (easier to have a GHA update/generate the file).

The main downside with variables as opposite to metadata are for conditional content, i.e., variables are not supported.

In the current states, metadata are great for `provider` because they work with the include shortcodes, so the same "code" can be depending on the parent document calling the file while variables are more for constant that does not need to change across documents.


Briefly:

- variables for constants
- metadata for "prose parameters"

cderv:
Interesting thanks for sharing user's view.

PR #2185, opened 2026-09-09
Document `quarto call axe`, the experimental site accessibility scanner

cwickham:
## Description

Adds `docs/websites/website-accessibility.qmd`, a user-facing page for the
experimental `quarto call axe` command (quarto-dev/quarto-cli#14815). 

Also updates `docs/output-formats/html-accessibility.qmd`. Its "Site-wide
accessibility checks" section ended with a paragraph about a mode we
"envision in the future". That paragraph now points at the new page.

## Decisions worth a look

- **The page is deliberately not in the sidebar.** It renders and it enters
  search and `llms.txt`, but an unlisted page under `docs/` gets no left
  sidebar and no breadcrumbs, so it reads full-width. That is the intended
  state while the command is experimental. Wiring it into the Websites
  section is a two-line change to `_quarto.yml`.
- **No guidance on excluding content in source.** The scanner's own generated
  README distinguishes two ledgers: inaccessibility an author chose as part
  of a lesson, versus a problem someone should eventually fix. Only the
  second belongs in the baseline. This page documents the baseline alone and
  says nothing about the first case, because the exclude-in-source pathway
  does not exist yet.
- **The page complements the generated `_axe-checks/README.md` instead of
  repeating it.** The README owns the field-by-field reference for a finding
  and for a baseline entry, and it is regenerated on every scan, so it always
  matches the build that wrote it. The page documents the workflow and links
  to it.
- **The worked example omits one finding.** A scan of the example site also
  reports `region` on `#quarto-skip-link`, on the home page only. That is a
  Quarto bug, not something the example's content causes, and it is filed as
  quarto-dev/quarto-cli#14875. The excerpts leave it out so that regenerating them after the
  fix reproduces the surrounding text unchanged. Everything else in them is
  exactly what the scan wrote, against Quarto 1.11.4.

## Follow-up in quarto-cli

Once this page has a URL, two places in quarto-cli need it (both currently
point at `dev-docs/axe-scan.md`):

1. `news/changelog-1.11.md`, the #14815 entry.
2. `src/command/call/axe/cmd.ts`, the command `.description()`.

The URL is `https://prerelease.quarto.org/docs/websites/website-accessibility.html`
until 1.11 ships.


cderv (approved):
Really great ! Thank you !! 

And it seems the screenshot tooling worked for you and 🤖 - cool !

PR #2186, opened 2026-09-11
docs: relocate book-vs-website key note so it is reachable from the Analytics anchor

cderv:
quarto-dev/quarto-cli#14879 reported that `book: plausible-analytics` has no effect in a book project. The underlying config bug is fixed in quarto-dev/quarto-cli#14881, but the report also pointed at a real docs gap: the rule that book projects use `book:` instead of `website:` for these options was already documented in the book-vs-website key note, just not where a reader lands. The note sat under Social Metadata while Analytics is roughly 150 lines further down, so a reader arriving directly at the Analytics anchor never saw it.

This moves the note to a page-level intro so its generalizing sentence ("substitute `book` for `website`") actually holds for the whole page as written, and adds a short pointer inside Analytics for readers who land there via a deep link rather than reading top-down. The book-vs-website example also gains a `plausible-analytics` key so it concretely covers the reported case.

This note is accurate from 1.11 onward (the fix in quarto-dev/quarto-cli#14881), so it targets `prerelease` rather than `main`.

Related to quarto-dev/quarto-cli#14879, quarto-dev/quarto-cli#14881.

### DIFF
--- a/docs/books/_book-vs-website-key.qmd
+++ b/docs/books/_book-vs-website-key.qmd
@@ -1,10 +1,11 @@
-One important thing to note about using website tools is that while these tools are added to websites within the `website` key, in a book you should include the same options in the `book` key. For example, in a website you would include a favicon and twitter card as follows:
+One important thing to note about using website tools is that while these tools are added to websites within the `website` key, in a book you should include the same options in the `book` key. For example, in a website you would include a favicon, twitter card, and analytics as follows:
 
 ```yaml
 website:
   favicon: logo.png
   twitter-card: true
   site-url: https://example.com
+  plausible-analytics: XXXXXX
 ```
 
 In a book you'd use the `book` key instead:
@@ -14,5 +15,6 @@ book:
   favicon: logo.png
   twitter-card: true
   site-url: https://example.com
+  plausible-analytics: XXXXXX
 ```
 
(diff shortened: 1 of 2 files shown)


docs/books/_book-vs-website-key.qmd
```diff
@@ -1,10 +1,11 @@
-One important thing to note about using website tools is that while these tools are added to websites within the `website` key, in a book you should include the same options in the `book` key. For example, in a website you would include a favicon and twitter card as follows:
+One important thing to note about using website tools is that while these tools are added to websites within the `website` key, in a book you should include the same options in the `book` key. For example, in a website you would include a favicon, twitter card, and analytics as follows:
 
 ```yaml
 website:
   favicon: logo.png
   twitter-card: true
   site-url: https://example.com
+  plausible-analytics: XXXXXX
```
cderv:
```suggestion
  plausible-analytics:
    path: _plausible_snippet.html
```

docs/books/_book-vs-website-key.qmd
```diff
@@ -14,5 +16,6 @@ book:
...
```
cderv:
```suggestion
  plausible-analytics:
    path: _plausible_snippet.html
```

PR #2188, opened 2026-09-15
Add typst-equations extension listing

jfsalzmann:
Adds a listing for typst-equations, a filter extension for authoring math formulas in native Typst syntax, rendering correctly in both HTML and Typst/PDF output from the same source. HTML output is achieved via SVG rendering of the Typst equations, with caching in place. Can be added as a global filter across any format. Fallback mechanisms for LaTeX or other outputs are in place.

Also see the extension's [repo](https://github.com/jfsalzmann/quarto-typst-equations).


cderv (commented):
Thanks for the submission.

This fills a real gap: `typst-math` (already listed) only covers `typst` output, and `typst-render`'s HTML math mode is still marked experimental upstream, so a stable SVG-based path for HTML is useful. 
Though are you aware of those existing extensions ? 

It would be worth noting somewhere in the description or README how this relates to `typst-math` and `typst-render`, since all three now sit in this same listing file solving adjacent problems. Could you add a line on that? 

Thank you

jfsalzmann:
Thanks Christophe, I have updated the README accordingly (and I'm indeed aware of the other two extensions - I think they have slightly different scopes & functionalities). I believe the respective clarification/comparison can better live in the README than the short description.

mcanouil:
Nice to see more Typst maths in Quarto.
For clarity only, since the three extensions now sit next to each other in the listing.

`math: typst` in `typst-render` also covers HTML.
It compiles each equation to native MathML through the Typst HTML export, and Quarto equation numbering and `@eq-` cross-references keep working.
The experimental part is the HTML export of Typst itself, which needs Typst 0.15 or later.

Below that version `typst-render` gives the equation back to the Quarto renderer, and it needs Quarto 1.10.18 or later.
The SVG path of `typst-equations` therefore covers toolchains that `typst-render` does not currently.

There is also a route that gives Typst maths in all output formats, and it needs neither `math: typst` nor Typst 0.15.
An inline `{typst}` expression such as `` `{typst} $pi r^2$` `` compiles to a cropped image that sits on the text baseline, and a `{typst}` code block does the same for display maths.
The same source therefore gives maths in HTML, Typst, LaTeX PDF, DOCX, and Reveal.js, with an `alt` attribute on the image.
Inline expressions in PowerPoint are the one exception, because Pandoc cannot put an image inside a text run.
Blocks work there.

Documentation: https://m.canouil.dev/quarto-typst-render/


Side note/FYI: your extension will be automatically added to <https://m.canouil.dev/quarto-extensions/> next week-end as retrieved from this repository.

